PR #45, opened 2023-05-30
docs: fixed broken radix links

AdrianGonz97:
Fixes Radix links for `alert-dialog`, `hover-card`, and `radio-group`.

The lack of `-`'s may be an error on Radix's end since their route for `aspect-ratio` (https://www.radix-svelte.com/docs/aspect-ratio) seems to follow this convention. We can update this again once that's sorted on their end.


huntabyte (approved):
Looks good! Thank you!

PR #68, opened 2023-06-04
Update hover-card.md

SkydiverTyler:
React -> Svelte :)


huntabyte:
Nice catch! Going to switch it up a bit since _technically_ it wasn't created by vercel!

Thank you!

Issue #125, opened 2023-06-24
Dependencies not compatible with Svelte 4


gursheyss:
### Current Behavior

When installing some components, such as labels or radio groups, 

```
Installing 1 component(s) and dependencies...
⠼ Label...file:///Users/gursheys/.npm/_npx/6326117962061f1d/node_modules/execa/lib/error.js:59
                error = new Error(message);
                        ^

Error: Command failed with exit code 1: npm install radix-svelte
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: download-aio@0.0.1
npm ERR! Found: svelte@4.0.0
npm ERR! node_modules/svelte
npm ERR!   dev svelte@"^4.0.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer svelte@"^3.58.0" from radix-svelte@0.8.0
npm ERR! node_modules/radix-svelte
npm ERR!   radix-svelte@"^0.8.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! 
npm ERR! For a full report see:
npm ERR! /Users/gursheys/.npm/_logs/2023-06-24T08_10_28_388Z-eresolve-report.txt
```

### Expected Behavior

Successfully added component and install radix-svelte

### Steps To Reproduce

`npx shadcn-svelte add input`

### More Information

Ive tried using `npm install radix-svelte --legacy-peer-deps` but adding some components still doesn't work.

huntabyte:
I'm talking to the Radix Svelte team about upgrading to Svelte 4.0 which will resolve these issues. We're at the mercy of that project at the moment. If you want to use the project immediately, I'd suggest using the latest Svelte 3.X release until this has been resolved.

huntabyte:
Additionally, [this PR](https://github.com/lucide-icons/lucide/pull/1388) will address the issues with Lucide.

mpost:
How does the retirement of the radix svelte library factor into resolving this issue?

https://github.com/radix-svelte/radix-svelte/blob/develop/ANNOUNCEMENT.md

Also how does it affect the future of this library, and possible migration to melt ui?

jasongitmail:
[...]

PR #132, opened 2023-06-26
add warnings for svelte v4 support

huntabyte:


huntabyte:
Add warnings to address #125

PR #140, opened 2023-07-08
Badge small improvements

trgsv:
- Add ring-offset-background to Badge
- Remove duplicate `badgeVariants`


huntabyte:
Thank you! Is the `ring-offset-background` in the original?

trgsv:
I think not. But this element is only focusable when it is an `<a>`. This is not part of the original but part of shadcn-svelte. 
I used it and noticed that I get an ugly white offset.

huntabyte:
Hey @trgsv, this should now be aligned with the original after the latest big overhaul!

Issue #148, opened 2023-07-31
Confusing installation instructions


Creative-Difficulty:
### Change Type

Correction

### Proposed Changes

In the installation docs (https://www.shadcn-svelte.com/docs/installation) under "New Project" at step 1 ("Run the CLI") it says "Use the SvelteKit CLI to create a new project. Support for Svelte v4 is coming soon, we are just waiting for our dependencies to be updated.", however the command below it uses svelte(-kit) >v4 (npm create svelte@v4.2.0 my-app). This is very confusing, also in the project setup, when running "npm create svelte@v4.2.0 my-app" a warning appears, saying trhat svelte >v4 isnt supported.

huntabyte:
Closed by #150

Issue #145, opened 2023-07-15
For Label, the Usage section doesn't match the Preview


jasongitmail:
### Change Type

Cleanup (formatting, typos, etc.)

### Proposed Changes

The Preview is "Accept terms and conditions".

The Usage shows: "Your email address"

For consistency, I assume these are intended to match.

huntabyte:
They are actually not intended to match! Usage is more of a general quick overview of component use, whereas the example component's code is under the 'Code' tab at the top!

Issue #144, opened 2023-07-15
Badge has text selection cursor, should be default


jasongitmail:
### Current Behavior

Hovering the badge shows the text selection cursor.

### Expected Behavior

Hovering the badge should show the default cursor (black arrow).

### Steps To Reproduce

1. Hover the badge

huntabyte:
Closed by #150

Issue #127, opened 2023-06-25
Uncaught TypeError in Slider component


trgsv:
### Current Behavior

Error occurs in several browsers.

### Expected Behavior

.

### Steps To Reproduce

1. Go to: https://www.shadcn-svelte.com/docs/components/slider
2. Open browser console
3. Move slider

AdrianGonz97:
Strange, it seems like this bug came back #5

AdrianGonz97:
It seems to be a Radix bug https://github.com/huntabyte/shadcn-svelte/issues/5#issuecomment-1565864812. As we'll be switching over to MeltUI soon, I expect this _may_ go away. We can keep this issue open in the mean time.

trgsv:
MeltUI sounds very good. Looking forward to that!

huntabyte:
Closed by #150

PR #150, opened 2023-08-02
🚀  Migrate Components to Melt UI

huntabyte:
Update (2023/08/15) - We’re wrapping up the CLI (thanks @AdrianGonz97) and we’ll be ready for takeoff. The headless primitives inherit just about all behavior/options from Melt with a few minor qualify of life differences when working with components. I still need to create a doc site for those but won’t have the community waiting before we publish these changes.

——

To reduce a lot of the unnecessary boilerplate & provide a better overall DX, I've decided to wrap the Melt UI builders in "primitives components", which will be similar to Radix UI, and will be published as the core dependency of `shadcn-svelte`. 

I know I know, "We tried that and it didn't work!", but that's only because we didn't have Melt under the hood.

[huntabyte/primitives](https://github.com/huntabyte/primitives)

I'll also track the status of each here so that you have a general idea of how far along we are.

Note: These are only the components that rely on a headless lib, the others are fine :)

## Things:
- [x] Migrate Primitives to Melt
- [x] Themes
- [x] Data Table
- [x] Updated CLI

Closes: #160 #125 #154 #148 #145 #144 #114 #127 #14


acidlake:
keep the great work Hunter, we are waiting on this, this is going to be missing ui library that svelte/kit was waiting

acidlake (commented):
this is working great, it looks great!!
Screenshot 2023-08-11 at 12 14 38 PM

spectrapulse:
Not to be rude or anything but aren't we missing #106 and #102?

PR #152, opened 2023-08-05
Add tabs

TGlide:
I'm following some different conventions:

- I didn't create a context getter for each and every element, as some people may want to get more things from the createBuilder method. Instead of having to define it in two places, i'd argue we just give them the whole context always
- I'm not passing down the value store or the onChange method, instead, I'm only allowing the defaultValue prop, but aliasing it to value. I then track changes so it can be used with bind:value. I think this is more fitting of a component-like structure
- Probably something else that i can't remember


huntabyte:
Closing as implemented in primitives library :)

Thanks for the 500IQ method @TGlide ❤️

PR #169, opened 2023-08-19
Update docs to point to shadcn-svelte

michal-weglarz:
Updated the docs to replace `shadcn-ui@latest` with `shadcn-svelte@latest`.


AdrianGonz97:
Thank you!

PR #174, opened 2023-08-19
Fix the shortcut in the menubar example

michal-weglarz:
It's a super tiny fix, but I noticed the usage example doesn't work out of the box when copied from the docs to my project.


huntabyte:
Thanks for this!

PR #203, opened 2023-08-22
Docs: change className to class prop

leonardsimonse:
The docs contain an example of a Select component. This example uses a non-existent className prop, which is changed to the class prop in this pull request.


huntabyte:
Thank you!

PR #258, opened 2023-09-04
fix: data table source link

jperchoc:
Fixed the data-table link so it does not target a 404 page anymore


huntabyte:
Thanks for this @jperchoc !

PR #284, opened 2023-09-09
Remove unneeded SvelteKit alias

jasongitmail:
I've found the 2nd version to be superfluous and all Shadcn components work without it. This is an opportunity to simplify for new users.

(Actually, the entire step could be removed because SvelteKit's default `$lib` alias without any changes to `alias: {}`, but I imagine specifying it gives explicit guidance to devs in case they overrode SvelteKit's default, so I didn't PR that.)


huntabyte:
Thanks @jasongitmail!

PR #297, opened 2023-09-15
fix: Passing ID to "Actions" column (docs)

jeffpohlmeyer:
## Issue
An `undefined` value  was being passed in as props to the "Actions" column in the "Update columns definition" section. This is because the `accessor` was _only_ passing in the email value so the `cell` method had no knowledge of the ID of the associated item. Also, the ID value lives on the `item.value` not the `item`.

### DIFF
--- a/apps/www/src/content/components/data-table.md
+++ b/apps/www/src/content/components/data-table.md
@@ -535,10 +535,10 @@ Now that we've defined our actions component, let's update our `actions` column 
       }
     }),
     table.column({
-      accessor: (item) => item,
+      accessor: "id",
       header: "",
       cell: ({ value }) => {
-        return createRender(DataTableActions, { id: value.id });
+        return createRender(DataTableActions, { id: value });
       }
     })
   ]);
@@ -620,10 +620,10 @@ Next, we'll add pagination to our table
       }
     }),
     table.column({
-      accessor: ({ email }) => email,
+      accessor: "id",
       header: "",
-      cell: (item) => {
-        return createRender(DataTableActions, { id: item.id });
+      cell: ({ value }) => {
+        return createRender(DataTableActions, { id: value });
       }
     })
   ]);
@@ -754,10 +754,10 @@ Let's enable the `addSortBy` plugin and import the icon we'll use to indicate th
       }
     }),
     table.column({
-      accessor: ({ email }) => email,
+      accessor: "id",
       header: "",
-      cell: (item) => {
-        return createRender(DataTableActions, { id: item.id });
+      cell: ({ value }) => {
+        return createRender(DataTableActions, { id: value });
       },
       plugins: {
         sort: {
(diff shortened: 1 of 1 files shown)


jeffpohlmeyer:
FWIW @huntabyte I did a little screen recording showing the effect with the current implementation in the docs and my proposed change. I was at first just recreating the tables in the docs (what would be really cool would be `REPL`s associated with each item in the docs but that would be a hell of a lot of work) to try to get a feel for how Data Tables work.

https://youtu.be/iOn79iu8uHE

huntabyte:
Hey @jeffpohlmeyer thanks for this! Something I recently discovered is that I'm doing things incorrectly in the data table, specifically for the select/checkbox column, as well as the actions column. Those columns should be using `table.display()` instead of `table.column` (see [here](https://svelte-headless-table.bryanmylee.com/docs/api/create-columns#table-display-displaydef-displaycolumn)).

If you're up for refactoring the table to use that (which prevents some weird bugs that can occur when hiding/filtering) then feel free! Otherwise I will update it soon as I have some time!

jeffpohlmeyer:
Sounds good. I'll definitely give it a shot but probably won't have a chance before Monday.

jeffpohlmeyer:
> Hey @jeffpohlmeyer thanks for this! Something I recently discovered is that I'm doing things incorrectly in the data table, specifically for the select/checkbox column, as well as the actions column. *Those columns should be using `table.display()` instead of `table.column` (see [here](https://svelte-headless-table.bryanmylee.com/docs/api/create-columns#table-display-displaydef-displaycolumn))*.
> 
> If you're up for refactoring the table to use that (which prevents some weird bugs that can occur when hiding/filtering) then feel free! Otherwise I will update it soon as I have some time!

I'm not sure that the actions column should be using `table.display()` instead of `table.column()` because you actually need the id value to be passed into the `data-table-actions.svelte` component and https://svelte-headless-table.bryanmylee.com/docs/api/create-columns#table-display-displaydef-displaycolumn indicates that you use `display` for "non data-related information to be displayed"

huntabyte:
> I'm not sure that the actions column should be using `table.display()` instead of `table.column()` because you actually need the id value to be passed into the `data-table-actions.svelte` component and [svelte-headless-table.bryanmylee.com/docs/api/create-columns#table-display-displaydef-displaycolumn](https://svelte-headless-table.bryanmylee.com/docs/api/create-columns#table-display-displaydef-displaycolumn) indicates that you use `display` for "non data-related information to be displayed"

@jeffpohlmeyer I thought about that as well, but the way I interpreted that is this represents something specific/unique within the dataset. The issues start to become more obvious when you also want to display an `id` column for example, especially with filtering, other plugins, etc.

We can still destructure and access the `id` of the specific row it's being rendered for to pass as a prop to the component, but we aren't necessarily displaying it as a part of the data if that makes sense lol

jeffpohlmeyer:
The only issue I ran into, though, when using the `display` method instead of the `column` method was accessing the data. The library assigns its own ID value to each entry and the data exists in the `row.original.id` element, but when I tried using that I got a funky type error.

I'll play around a bit more in the morning (Eastern time)

huntabyte:
No worries @jeffpohlmeyer, I recently did it and was trying to find the branch but I guess I never committed it.

To get around that type of error, you have to narrow the typing using this function [here](https://svelte-headless-table.bryanmylee.com/docs/api/body-row#isdisplay-boolean).

So something like:

```ts
cell: ({ row }) => {
	if (row.isDisplay()) {
		// now I know the type and can access row.original
		console.log(row.original.id)
	}
}
```

PR #299, opened 2023-09-17
chore: remove tailwindcssAnimate from docs

sokyran:
During project installation process I have noticed `tailwindcssAnimate` was not found, so I have installed it. But later I have found this pull request #236 that removes `tailwindcssAnimate` from project so I believe this inaccuracy should be removed from `installation.md` as well


huntabyte:
Thanks for this, @sokyran !

PR #307, opened 2023-09-20
Fix disabled checkbox styles

### LINKED ISSUE #306: Disabled checkbox styles aren't working
### Current Behavior

The styles `disabled:cursor-not-allowed disabled:opacity-50` on the Checkbox are not being applied when the checkbox is disabled.

Adding the following styles seems to fix it, but I'm not sure if it's the proper way of doing it:
`data-[disabled=true]:cursor-not-allowed data-[disabled=true]:opacity-50`

### Expected Behavior

The disabled styles should be applied and the `not-allowed` cursor should be shown.

### Steps To Reproduce

Hover over the checkbox in the disabled example:
https://www.shadcn-svelte.com/docs/components/checkbox

_No response_

### More Information

_No response_

tmadge:
closes #306 

### Before:
image

### After:
image

### DIFF
--- a/apps/www/src/lib/registry/default/example/checkbox-disabled.svelte
+++ b/apps/www/src/lib/registry/default/example/checkbox-disabled.svelte
@@ -7,7 +7,7 @@
 	<Checkbox id="terms" disabled />
 	<Label
 		for="terms2"
-		class="text-sm font-medium leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
+		class="text-sm font-medium leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70 peer-data-[disabled=true]:cursor-not-allowed peer-data-[disabled=true]:opacity-70"
 	>
 		Accept terms and conditions
 	</Label>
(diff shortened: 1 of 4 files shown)


apps/www/src/lib/registry/default/example/checkbox-disabled.svelte
```diff
@@ -7,7 +7,7 @@
 	<Checkbox id="terms" disabled />
 	<Label
 		for="terms2"
-		class="text-sm font-medium leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
+		class="text-sm font-medium leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70 peer-data-[disabled=true]:cursor-not-allowed peer-data-[disabled=true]:opacity-70"
```
huntabyte, line 10: I think you should be able to do `peer-data-[disabled]:cursor-not-allowed`, etc., as if I remember correctly we don't set that property at all if it _isn't_ disabled.

huntabyte (changes requested):
Other than the small thing in the styles it looks great! Thank you!

apps/www/src/lib/registry/default/example/checkbox-disabled.svelte
tmadge: I just double checked and it looks like `data-disabled="false"` is added when not disabled.

image

huntabyte: Interesting, that shouldn't be the case but will work it out on the Melt side. In the meantime, this looks good to merge, thanks for this!

Issue #275, opened 2023-09-07
Mistake in Table guide


philippviereck:
https://github.com/huntabyte/shadcn-svelte/blob/22bd64cf3f03ce2b65984686164d060936700a4d/apps/www/src/content/components/data-table.md?plain=1#L623-L627

```
table.column({
			accessor: ({ id }) => id,
			header: '',
			cell: ({ value }) => {
				return createRender(DataTableActions, { id: value });
			},
			// ...
		})
```

huntabyte:
Yeah, I've actually been meaning to update the data table guide as after I published it I discovered that for the columns that aren't represented by the data, in this case, the select checkbox and the actions, we should be using `table.display` instead of `table.column`.

If anyone wants to update it, feel free to open a PR, otherwise, I will update it when I have some time!

philippviereck:
on it, but I have to admit, it's proving to be more challenging than I initially anticipated! 😅

philippviereck:
I just found: https://github.com/huntabyte/shadcn-svelte/pull/297. Let me know how to continue.

PR #322, opened 2023-09-25
fix: Mistake in Table guide #275

philippviereck:
fixes: #275 

- I updated the docs and the svelte-components I found to use: `Table.Display` instead of `Table.Column` where needed.
- I also updated the line-numbers for the code highlighting as well as minor adaptions in the text as a result of the code changes.
- I changed `export const payments: Payment[] = [` to `export const data: Payment[] = [` to be consistent with the rest of the guide.
- I moved `type Payment` 

NOTE:
I'm not really sure if this is the proper way to access and pass the data as a prop, maybe @bryanmylee can double check:
```
cell: ({ row }, { data }) => {
        const payment = derived(data, ($data) => $data[Number(row.id)]);
        return createRender(DataTableActions, {
          payment
        });
      }
```

Also, now that `Table.Display` is used instead of `Table.Column` there's this weird behaviour that when hiding the email column all rows disappear see: https://github.com/bryanmylee/svelte-headless-table/issues/160

 feedback and review on these changes are greatly appreciated

### DIFF
--- a/apps/www/src/content/components/data-table.md
+++ b/apps/www/src/content/components/data-table.md
@@ -68,7 +68,7 @@ type Payment = {
   email: string;
 };
 
-export const payments: Payment[] = [
+export const data: Payment[] = [
   {
     id: "728ed52f",
     amount: 100,
@@ -114,14 +114,16 @@ Let's start by building a basic table.
 Before we can initialize a table, we need to get our data. You can retrieve your data from anywhere, but for this example we'll use a `payments` array.
 
 ```svelte showLineNumbers title="routes/payments/data-table.svelte"
-<script lang="ts">
-  type Payment = {
+<script lang="ts" context="module">
+  export type Payment = {
     id: string;
     amount: number;
     status: "pending" | "processing" | "success" | "failed";
     email: string;
   };
+</script>
 
+<script lang="ts">
   const data: Payment[] = [
     {
       id: "m5gr84i9",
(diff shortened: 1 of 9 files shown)


philippviereck:
I just found: https://github.com/huntabyte/shadcn-svelte/pull/297. Let me know how to continue.

huntabyte:
Let's see what @jeffpohlmeyer has to say, maybe you could wrap up his PR if he doesn't have the time, either way I appreciate the contribution!

jeffpohlmeyer:
I've been busy with work for the past week and a half and haven't had a chance to update mine. I'll take a look in the morning but if this addresses the shortcomings of mine then I'm happy to close my PR

philippviereck:
I think my PR addresses everything @jeffpohlmeyer.

But before merging these should be solved/agreed upon (see my initial comment):
- Best practice to access / pass data when using `table.display()`
- https://github.com/bryanmylee/svelte-headless-table/issues/160

huntabyte:
Hey @philippviereck,

You should be able to get the row's ID using some type narrowing helper functions provided by the library,  [here](https://svelte-headless-table.bryanmylee.com/docs/api/body-row#isdisplay-boolean).

So something like:

```ts
cell: ({ row }) => {
	if (row.isDisplay()) {
		// now I know the type and can access row.original
		return createRender(DataTableActions, {
          payment: row.original.id
        });
	}
}
```

For that issue you tagged, I remember that being a pain in the ass for me as well when building it out. I want to say it has something to do with the filter or sort, and ensuring they are disabled for all columns that don't need them. I was really hoping that switching to `table.display` would fix it for some reason, but I guess that's not the case.

PR #326, opened 2023-09-29
Fix backdrop blur support rule

ollema:
`supports-backdrop-blur:bg-background/60` does nothing

`supports-[backdrop-filter]:bg-background/60` does something!

see: https://github.com/shadcn-ui/ui/pull/1628


huntabyte:
Thanks for this @ollema !

huntabyte (approved):

PR #328, opened 2023-09-30
Use regular click event

ollema:
I believe the `m-*` events are not longer used in melt/bitsui/shadcn-svelte?


huntabyte:
Thank you!

PR #341, opened 2023-10-10
fx: add the special `<alpha-value>` placeholder

### LINKED ISSUE #333: Current implementation of `hsl` doesn't inject `alpha value` when using an opacity modifier
### Current Behavior

Currently, if I wanted to style something using opacity modifiers, it would not work.

For example:
```svelte
<button class="... border-primary/20 bg-primary/10 ...">
	Reduced Opacity
</button>
```

If you do that, then the semblance of reduced opacity isn't applied.

### Expected Behavior

I should be able to control the opacity of colours using the colour opacity modifier.

e.g.
```svelte
<button class="bg-sky-500/100 ..."></button>
<button class="text-sky-500/75 ..."></button>
<button class="border-sky-500/50 ..."></button>
```

### Steps To Reproduce

1. Pick any component whatsoever, say, a button.
2. Try giving it a background of `bg-indigo-500/50`.
3. That style won't be applied.

I tried implementing a fix locally, and all worked as expected.
The fix is relatively straight forward, and I volunteer to work on this issue.

_No response_

### More Information

The reason is that the current definition of colours in the configuration file doesn't inject the special `<alpha-value>` placeholder that Tailwind will use to inject the alpha value when using an opacity modifier

shyakadavis:
Fixes #333 

Per the TailwindCSS docs [here](https://tailwindcss.com/docs/customizing-colors#using-css-variables), when defining your colors in your configuration file, you need to be sure to include the color space you’re using, and the special `<alpha-value>` placeholder that Tailwind will use to inject the alpha value when using an opacity modifier.

This is useful when controlling the opacity of colours using the colour opacity modifier.
e.g. `bg-indigo-500/50`

### DIFF
--- a/apps/www/src/content/installation.md
+++ b/apps/www/src/content/installation.md
@@ -146,38 +146,38 @@ const config = {
     },
     extend: {
       colors: {
-        border: "hsl(var(--border))",
-        input: "hsl(var(--input))",
-        ring: "hsl(var(--ring))",
-        background: "hsl(var(--background))",
-        foreground: "hsl(var(--foreground))",
+        border: "hsl(var(--border) / <alpha-value>)",
+        input: "hsl(var(--input) / <alpha-value>)",
+        ring: "hsl(var(--ring) / <alpha-value>)",
+        background: "hsl(var(--background) / <alpha-value>)",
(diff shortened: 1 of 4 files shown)


huntabyte:
Thank you @shyakadavis !

Issue #338, opened 2023-10-08
`<Input type="file" />` has unreadable text in dark mode


phenomen:
### Current Behavior

`<Input type="file" />` has a black "Choose file" label on a black background, making it unreadable.

![chrome_TchH7pw6KC](https://github.com/huntabyte/shadcn-svelte/assets/138311/b0907d5e-e6ca-4d0e-a37b-d599b8ee60eb)

### Expected Behavior

Here is how it looks in `shadcn-ui/ui`:

![image](https://github.com/huntabyte/shadcn-svelte/assets/138311/5326fdc2-7c32-4386-85d6-68c44b7742b6)

### Steps To Reproduce

1. https://www.shadcn-svelte.com/docs/components/input#file
2. Toggle dark mode

HubbeDev:
@phenomen @huntabyte Looks fine on my end. Tested in Arc and safari.

<img width="661" alt="file-input-dark-mode" src="https://github.com/huntabyte/shadcn-svelte/assets/57641199/b3e6c422-4788-44bb-9699-8b17b0f7f406">

huntabyte:
What browser are you using @phenomen?

phenomen:
> What browser are you using @phenomen?

Google Chrome 118
Can still reproduce it here https://www.shadcn-svelte.com/docs/components/input#file

fen89:
> > What browser are you using @phenomen?
> 
> Google Chrome 118 Can still reproduce it on `shadcn-svelte` docs here https://www.shadcn-svelte.com/docs/components/input#file Not reproducible on the original `shadcn/ui` https://ui.shadcn.com/docs/components/input#file
> 
> And it's not just docs, it happens in my app with `shadcn-svelte` too.

I can confirm this behavior, happen on my side as well (reproduced in  Chrome v117, Firefox v118) :

 Chrome v117

![file-dark-mode](https://github.com/huntabyte/shadcn-svelte/assets/4036376/1a6e251d-7061-4c5e-9f6c-3c45da0675e7)

If we remove all styles from the inputs on both shadcn-svelte and original shadcn, they both differ:

```html
<input class="" id="picture" type="file">
```

in original shadcn:
![image](https://github.com/huntabyte/shadcn-svelte/assets/4036376/929e893e-6841-43d8-9b81-47a12b925078)

in shadcn-svelte:
[...]

PR #353, opened 2023-10-17
Fix file input button text not visible in darkmode

HubbeDev:
**Fix for #338** 

Added foreground text color to file button.


huntabyte:
Thank you! @HubbeDev

huntabyte (approved):

PR #371, opened 2023-10-25
Fix: data-highlight for select

saturnonearth:
Add highlight on hover for `Select` component, and switch over to using `data-[highlighted]` instead of `focus:`.

Issue that still remains:

CleanShot 2023-10-25 at 12 42 38@2x

`cursor: 'pointer'` being injected somewhere on the `Select` items, overriding the `cursor-default` class that is already present.


huntabyte:
Changesets are only needed when changes are made to the cli @saturnonearth 😃

apps/www/src/lib/registry/default/ui/select/select-item.svelte
```diff
@@ -18,7 +18,7 @@
 	{disabled}
 	{label}
 	class={cn(
-		"relative flex w-full cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+		"relative flex w-full cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none hover:bg-accent hover:text-accent-foreground data-[highlighted]:bg-accent data-[highlighted]:focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
```
huntabyte, line 21: We shouldn't need `hover` classes here.

You also forgot to remove `focus` from one of the highlighted selectors 😃
should be - `data-[highlighted]:text-accent-foreground`

apps/www/src/lib/registry/new-york/ui/select/select-item.svelte
```diff
@@ -18,7 +18,7 @@
 	{disabled}
 	{label}
 	class={cn(
-		"relative flex w-full cursor-default select-none items-center rounded-sm py-1.5 pl-2 pr-8 text-sm outline-none data-[highlighted]:bg-accent data-[highlighted]:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+		"relative flex w-full cursor-default select-none items-center rounded-sm py-1.5 pl-2 pr-8 text-sm outline-none hover:bg-accent data-[highlighted]:bg-accent hover:text-accent-foreground data-[highlighted]:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
```
huntabyte, line 21: Same here, I don't think we need hover classes as highlighted should handle hover events and it can be weird if you leave your mouse on one item and then use your keyboard to navigate down, it will technically show two items as highlighted.

huntabyte (changes requested):
Just a few minor things

apps/www/src/lib/registry/new-york/ui/select/select-item.svelte
saturnonearth: > Same here, I don't think we need hover classes as highlighted should handle hover event

Yea that makes sense, but without hover mouseover it does not highlight on hover - but I am not sure if that is because melt-ui needs to be updated or something else?

>  It can be weird if you leave your mouse on one item and then use your keyboard to navigate down, it will technically show two items as highlighted.

I agree, assuming highlighted works in both cases, that's fine. But if we _do_ need to use `hover:` I don't think it would be a big deal as  I _assume_ (possibly incorrectly) someone using a keyboard to navigate would be less likely to also be using the mouse for navigation.

huntabyte:
This issue/PR is blocking this from working as we'd expect: https://github.com/melt-ui/melt-ui/issues/667 hopefully gets merged and released soon and I'll merge this in once we confirm it works as expected!

huntabyte (approved):

saturnonearth:
omg can't believe I forgot the bg class today is not my day 😅

Issue #392, opened 2023-11-01
Card.Title tag type does not have value `p` in type `HeadingLevel`


Skylli202:
### Change Type

Correction

### Proposed Changes

In this section of the documentation, [Modify the heading level](https://www.shadcn-svelte.com/docs/components/card#modify-the-heading-level) there is the following example:

```svelte
<Card.Title tag="p">This will render a P tag</Card.Title>
```

Which in my editor, give me the following error `Type '"p"' is not assignable to type 'HeadingLevel | undefined'` as the screenshot below show.

![image](https://github.com/huntabyte/shadcn-svelte/assets/38432216/68827d9f-9df9-45e4-ba3b-e78d5ac08608)

Which is expected as the type definition of `HeadingLevel` is the following union type: `export type HeadingLevel = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';` defined in [`bits-ui/src/components/ui/card/index.ts`](https://github.com/huntabyte/bits-ui/blob/main/src/components/ui/card/index.ts)

I am not knowledgeable enough to know if the appropriate fix is:
- Remove this example as it is incorrect
- Add `"p"` in the union type of `HeadingLevel`

I'd be really happy to do a PR with the appropriate fix once the appropriate fix has been decided by maintainers. And if it is the latter, which file in which repo to do the PR.

My gut is telling me that it the correct fix is to edit the `HeadingLevel` type as the component is rendering a paragraph tag in the DOM, if so it is then just a TypeScript typing error.

huntabyte:
Hey @Skylli202 thanks for pointing that out! Oversight on my side!

I don't think we should add `'p'` to the `HeadingLevel` type, instead the example should be updated to replace `'p'` with a valid heading. It will still work of course, but you'll get a type error which makes you stop and think is this really what I want to do?

If so, you're more than welcome to modify the `HeadingLevel` in your local component, but from the library side I think we try to nudge peeps into using solid semantics, a card heading is in fact a `heading`, so it should be tagged appropriately.

Skylli202:
I do agree on the semantics side. I'll try to push a PR on the documentation if you have not already done it.

huntabyte:
Closed by #393

PR #393, opened 2023-11-02
Fix: <Card.Title> example with `tag="p"`

Skylli202:
For context, see: #392

- I removed the example with `p` as it is semantically false.
- I changed the examples `tag="h2"` and `tag="h5"` to `tag="h1"` and `tag="h6"` as they are the upper and lower limit of heading.
- Added `...` to show that the property `tag` accept all value in-between the first and the second example.

### DIFF
--- a/apps/www/src/content/components/card.md
+++ b/apps/www/src/content/components/card.md
@@ -55,15 +55,13 @@ By default, the `<Card.Title>` component renders an `<h3>` element. You can chan
 For example:
 
 ```svelte
-<Card.Title tag="h2">This will render an H2</Card.Title>
+<Card.Title tag="h1">This will render an H2</Card.Title>
 ```
 
-```svelte
-<Card.Title tag="h5">This will render an H5</Card.Title>
-```
+...
 
 ```svelte
-<Card.Title tag="p">This will render a P tag</Card.Title>
+<Card.Title tag="h6">This will render an H5</Card.Title>
 ```
 
 ## Examples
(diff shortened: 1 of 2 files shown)


Skylli202:
Result on local dev looks like this:
image

Skylli202:
I'll read about how to add a changeset in a couple hours, this is brand new to me and I do not have the time right now.

huntabyte:
No need to add a changeset unless we're changing to CLI, since thats the only thing being distributed via npm! I just deleted it! @Skylli202

apps/www/src/content/components/card.md
```diff
@@ -55,15 +55,13 @@ By default, the `<Card.Title>` component renders an `<h3>` element. You can chan
...
 ```
 
-```svelte
-<Card.Title tag="h5">This will render an H5</Card.Title>
-```
+...
 
 ```svelte
-<Card.Title tag="p">This will render a P tag</Card.Title>
+<Card.Title tag="h6">This will render an H5</Card.Title>
```
huntabyte, line 64:
```suggestion
<Card.Title tag="h5">This will render an H5</Card.Title>
```

huntabyte (changes requested):

apps/www/src/content/components/card.md
Skylli202: Hi, I do not understand why you requested these changes. But it made me see that I changed the `tag`, but did not change the text of the example. I'm pushing a new commit as we speak.

huntabyte (approved):
Thanks a ton for this! @Skylli202

PR #405, opened 2023-11-04
Update command.md

jakobpesch:
Seems there is a small mistake in the import statement of the usage example of the command component.

It imports from `collapsible` instead of `command`.


huntabyte:
Whoops, thanks for this!

huntabyte (approved):

PR #444, opened 2023-11-13
Wrapped Select and Chevron in a div for NativeSelect component

### LINKED ISSUE #305: `<Form.NativeSelect>` doesn't properly render arrow svg
### Current Behavior

`<Form.NativeSelect>` appears to improperly position the down arrow for the select component absolutely on the page.

```svelte
<Form.Item>
	<Form.Label>Notify Employee</Form.Label>
	<Form.NativeSelect required class="w-full">
		<option value={null}>Select</option>
		{#each data.reminderRecipients as employee}
			<option value={employee.id}>{employee.lastName}, {employee.firstName}</option>
		{/each}
	</Form.NativeSelect>
	<Form.Validation />
</Form.Item>
```

Screenshot 2023-09-19 at 4 40 01 PM

### Expected Behavior

The down arrow should render inline with the native select component and not be absolutely positioned on the page.

### Steps To Reproduce

See sample code above.

_No response_

### More Information

_No response_

babakfp:
Fixes #305

### DIFF
--- a/apps/www/src/lib/registry/default/ui/form/form-native-select.svelte
+++ b/apps/www/src/lib/registry/default/ui/form/form-native-select.svelte
@@ -11,14 +11,16 @@
 	export { className as class };
 </script>
 
-<FormPrimitive.Select
-	class={cn(
-		buttonVariants({ variant: "outline" }),
-		"appearance-none bg-transparent font-normal",
-		className
-	)}
-	{...$$restProps}
->
-	<slot />
-</FormPrimitive.Select>
-<ChevronDown class="absolute right-3 top-2.5 h-4 w-4 opacity-50" />
+<div class="relative">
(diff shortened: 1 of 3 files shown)


babakfp:
Should I remove this `relative` class here?

image

image

https://www.shadcn-svelte.com/examples/forms/appearance

`src\routes\examples\forms\appearance\appearance-form.svelte`

I think this was added to solve the same issue discussed in #305.

babakfp:
> Yeah if we're already adding a relative container into it we _shouldn't_ need it. When you try removing it as long as everything still looks good for that example form I'm cool with dropping it.

All good. Added the [Removed class](https://github.com/huntabyte/shadcn-svelte/pull/444/commits/128619cca702b2333e1e5a069babf9339791e25f) commit.

PR #449, opened 2023-11-15
docs: add ESLint configuration section to installation docs

ollema:
with this commit, a `.eslintrc` file _should_ be added to the `.../components/ui` folder when initialising a new project with `npx shadcn-svelte@latest init`.

I write _should_ because I do not know how to test this 😅 so I have coded it "blind"

I added this because it's a common problem that people run into, this solution works without modifying existing project wide `.eslintrc.cjs` files.

a downside is that this is added without checking if the project uses `eslint` to begin with. a check for that could be added I guess!

feel free to close if you dislike this idea - no problem

### DIFF
--- a/apps/www/src/content/installation.md
+++ b/apps/www/src/content/installation.md
@@ -413,3 +413,29 @@ src
 - The `app.pcss` file contains the global CSS.
 
 That's it. You can now start adding components to your project.
+
+## ESLint configuration
+
+If you are using ESLint to find problems in your code, some components might trigger false positives depending on your ESLint configuration. For example, you could end up with lint errors when components define `$$Props` to specify the type for `restProps` because `$$Props` is not directly used in the rest of the component.
+
+To ignore these linting errors, you can modify your ESLint configuration.
+
+One option is to add a `.eslintrc` file in the directory where you define your components, for example `$lib/components/ui`:
+
+```json title="src/lib/components/ui/.eslintrc"
+{
+  "rules": {
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_",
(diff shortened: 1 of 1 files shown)


huntabyte:
Hey @ollema, I thought about this recently but ultimately decided against it, because as you mentioned, we have to make a lot of assumptions that are outside the scope of this project. We don't want to be in the business of modifying/setting ESLint configs on the project's behalf.

One thing we could do instead is add a section to the installation page of the docs, which would include something like 'Recommended ESLint Config' and include those pieces and _why_ they are recommended!

Will also loop in @AdrianGonz97 for his thoughts on this as well!

huntabyte:
Hey @ollema, are you wanting to add this to the docs or should we close this issue for now?

ollema:
> Hey @ollema, are you wanting to add this to the docs or should we close this issue for now?

sure, I can do that. I missed that suggestion in your original message

niemyjski:
We shouldn't be adding any eslint overrides to the components folder. We should fix any default rule errors in this code base.

huntabyte:
It's not an issue with this code base, it's an issue with the default eslint rules, which tbh will be resolved in Svelte 5 but in the meantime requires you to ignore the `$$Events` & `$$Props` if unused. Otherwise you looks intellisense/type safety for the HTMLAttributes props on certain elements that don't have other props.

Here's an eslint config you can use to ignore these in your codebase, but we're not going to be modifying/messing with anyone's eslint config.

https://gist.github.com/huntabyte/b73073a93a7a664f3cbad7c50376c9c9

ollema:
ready for review @huntabyte

huntabyte (approved):
Thanks man! Love the multiple options!

PR #452, opened 2023-11-15
feat: add label field on mobile menu

### LINKED ISSUE #450: Add the labels on the mobile menu
### Describe the feature in detail (code, mocks, or screenshots encouraged)

I want to add the label next to the links in the mobile menu like on the bigger screen (`aside` section). It should look like this on mobile

image

### What type of pull request would this be?

Docs

### Provide relevant links or additional information.

_No response_

NDOY3M4N:
This closes #450

### DIFF
--- a/apps/www/src/lib/components/docs/nav/mobile-nav.svelte
+++ b/apps/www/src/lib/components/docs/nav/mobile-nav.svelte
@@ -49,6 +49,13 @@
 										class="text-muted-foreground"
 									>
 										{item.title}
+										{#if item.label}
+											<span
+												class="ml-2 rounded-md bg-[#adfa1d] px-1.5 py-0.5 text-xs leading-none text-[#000000] no-underline group-hover:no-underline"
+											>
+												{item.label}
+											</span>
+										{/if}
 									</MobileLink>
 								{/if}
 							{/each}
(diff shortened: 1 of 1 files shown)


apps/www/src/lib/components/docs/nav/mobile-nav.svelte
```diff
@@ -49,6 +49,13 @@
 										class="text-muted-foreground"
 									>
 										{item.title}
+										{#if item.label}
+											<span
+												class="ml-2 rounded-md bg-[#adfa1d] px-1.5 py-0.5 text-xs leading-none text-[#000000] no-underline group-hover:no-underline"
```
leonardsimonse, line 54: Isn't `group-hover:no-underline` redundant as `no-underline` already has the same effect?

leonardsimonse (changes requested):

apps/www/src/lib/components/docs/nav/mobile-nav.svelte
NDOY3M4N: Oh you're right. I just copied the markup from `docs-sidebar-nav-items.svelte`. I'll remove these 2 classes here.

huntabyte: Whoops!

huntabyte (approved):
Thank you!

PR #454, opened 2023-11-16
Fix: Tasks example: render column header instead of id

leonardsimonse:
Tasks example renders column id in the column header, the header is a single word in the example, so the example doesn't show the error, but it doesn't render the actual column header. A header set to string 'Hello world' and an id set to 'helloWorld' would render as 'HelloWorld' without this change. This pull request fixes that.


huntabyte (approved):
Thank you!

PR #488, opened 2023-11-29
fix incorrect import and broken code in Data Table doc

SumitBando:
Data Table doc has incorrect import for Pagination, and hence broken.


huntabyte (approved):
Nice catch! Thanks for this!

PR #505, opened 2023-12-06
fix(cli): handle "pnpm option"

fernandolguevara:


AdrianGonz97:
This change isn't necessary as `ni` returns `pnpm` anyway. Though, it does raise the question as to why the `bun` case is present in the first place. I'll have to test it out later.

fernandolguevara:
@AdrianGonz97 In my case `ni` can not detect the underlying runtime bc lockfiles are not enabled on my project, I had to add `packageManager` property on `package.json`.

would be nice to input the cli what PM should use on the `init` process or via flag `--packageManager`

AdrianGonz97:
@fernandolguevara That's a separate problem that I've fixed before over on melt's CLI: https://github.com/melt-ui/cli/commit/35088a0f1e53e76ff82fad51e1bfd0ab0c916e8e

I'll apply the same fix to this repo as well! This should resolve your issue.

PR #526, opened 2023-12-08
chore: update mode-watcher & add fallback mode before stores load

ollema:
just wanted to update `mode-watcher`.

now with `mode-watcher` > `0.1.0`, `$mode` is (more accurately) `undefined` until the stores have loaded correctly.

this is usually not a problem for FOUC sensitive applications since those should rely on the presence of the `dark` class in the `html` element and use the `dark:` tailwind variant.

for less FOUC sensitive applications like the styling of these buttons, it should be perfectly fine to fallback to a certain theme (`"dark"` in this case). the fallback is not really needed but it's there so that `pnpm check` is happy

### DIFF
--- a/apps/www/package.json
+++ b/apps/www/package.json
@@ -81,7 +81,7 @@
 		"esm-env": "^1.0.0",
 		"formsnap": "^0.4.1",
 		"lucide-svelte": "^0.292.0",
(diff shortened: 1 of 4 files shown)


huntabyte (approved):
Thanks @ollema !

PR #527, opened 2023-12-09
docs: Added `DatePickerWithRange` to the Dashboad page

asalvi0:
Added `DatePickerWithRange` to the Dashboard page, matching the original shadcn-ui site.

Also, changed the style of some elements in the home page to match the rest.


huntabyte (approved):
Thank you!

Issue #531, opened 2023-12-10
Using Popover + Select (preventScroll) -> Date Picker with Presets Example


roostad:
### Current Behavior

When using Select (preventScroll set to true by default) inside a Popover (preventScroll set to false by default), there is an issue where overflow remains hidden if you click the Select component followed by clicking outside the area of the Popover component (no selection).

This is just a quick note for an adjustment on this in the Date Picker with Presets example: https://www.shadcn-svelte.com/docs/components/date-picker#with-presets

### Expected Behavior

To re-enable scrolling after clicking outside the Popover with the Select menu open, preventScroll should also be set on Popover.Root.

```
<Popover.Root openFocus preventScroll>
    ...
</Popover.Root>
```

Pull request added: https://github.com/huntabyte/shadcn-svelte/pull/532

### Steps To Reproduce

See link to example below.

### Link to Reproduction / Stackblitz ([reproduction template](https://shadcn-svelte.com/repro))

https://www.shadcn-svelte.com/docs/components/date-picker#with-presets

huntabyte:
Duplicate of #514 - its a bug and will be fixed once I release a patch update to bits tomorrow

PR #532, opened 2023-12-10
fix: Added preventScroll to Popover.Root in Date Picker with Presets example

roostad:
Addresses small issue described below (see also https://github.com/huntabyte/shadcn-svelte/issues/531)

When using Select (preventScroll set to true by default) inside a Popover (preventScroll set to false by default), there is an issue where overflow remains hidden if you click the Select component followed by clicking outside the area of the Popover component (no selection).

This is just a quick note for an adjustment on this in the Date Picker with Presets example: https://www.shadcn-svelte.com/docs/components/date-picker#with-presets

### DIFF
--- a/apps/www/src/lib/registry/default/example/date-picker-with-presets.svelte
+++ b/apps/www/src/lib/registry/default/example/date-picker-with-presets.svelte
@@ -26,7 +26,7 @@
 	];
 </script>
 
-<Popover.Root openFocus>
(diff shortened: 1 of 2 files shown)


huntabyte:
Hey there, this is actually being tracked and a fix will be live tomorrow for it with the latest patch release of bits-ui!

Thanks for taking the time to do this anyways!

PR #541, opened 2023-12-16
fix(cli): Handle tailwind.config.ts file properly in init

BlankParticle:
# What has changed
I have removed the check that checks if the specified tailwind config file already exists and changed the replace regex that is used to delete the `.cjs` file.

# Motivation for the change
I was trying out this and so I made a new sveltekit project and tried to init shadcn.
It gave me option to use `tailwind.config.ts` as a config file so I tried to use that. I faced an error that `tailwind.config.ts` doesn't exists. It's dumb as the file is going to be replaced anyways.
So I renamed the file to `tailwind.config.ts` and ran the command again.
It seemed to work and made the file but quickly deleted the file afterwards.

Turns out `"tailwind.config.ts".replace(".js",".cjs")` returns `tailwind.config.ts` anyways. So the new config gets deleted instead of `.cjs` one

This PR addresses these 2 issues.

### DIFF
--- a/.changeset/wicked-eels-clap.md
+++ b/.changeset/wicked-eels-clap.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix(cli): Handle tailwind.config.ts file properly in init
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -160,18 +160,7 @@ async function promptForConfig(
 			message: `Where is your ${highlight(
 				"tailwind.config.[cjs|js|ts]"
 			)} located?`,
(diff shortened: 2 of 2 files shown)


huntabyte:
Please follow the [Contribution Guidelines](https://github.com/huntabyte/shadcn-svelte/blob/main/CONTRIBUTING.md#how-to-contribute) prior to submitting a pull request.

BlankParticle:
Hello, I belive that the PR was close because I didn't make an Issue prior to submitting the PR.

I was going to make an Issue but saw that the Issue was fixed by #543, but that PR is also not made against any Issue. Can you please point out what I did wrong so that I don't repeat the same mistake in future

Thanks

Issue #298, opened 2023-09-16
How do i update to a newer version


Creative-Difficulty:
### Current Behavior

Updating is kind of a pain in the a$$, as i have to create a new branch, delete all files in the branch, reinstall svelte and shadcn etc. and then copy my routes and other files over from another branch.

### Expected Behavior

A command line flag when installing shadcn-svelte would be very useful, or an entirley new command in the CLI(`npx shadcn-svelte@latest init --update`, or `npx shadcn-svelte@latest update`).

### Steps To Reproduce

-

ftognetto:
I usually use `npx shadcn-svelte@latest add button --overwrite` (I took as an example the button component)
If I previously edited button component somehow (eg changing background color), check with git diff and redo the changes on the new installed component.
But I do agree It's a price to pay for having the components installed directly in your code instead of having an npm package. Also on original shadcn library there is the same problem

Creative-Difficulty:
Good solution, however, with this method the `utils.ts` is still not updated, which was the culprit of many errors and hours wasted in my case. An update would be nice @huntabyte

yomp-admin:
Yah I would also want this.

huntabyte:
We could possibly add an `update utils` which would overwrite your existing utils with the latest one, but that's as far as it'd go at this point in time.

Creative-Difficulty:
> We could possibly add an `update utils` which would overwrite your existing utils with the latest one, but that's as far as it'd go at this point in time.

That would be very cool

DePasqualeOrg:
I'm getting this error when I run the update command:

```
$ npx shadcn-svelte@latest update --all
Running the following command will overwrite existing files.
Make sure you have committed your changes before proceeding.


FetchError: request to https://shadcn-svelte.com/registry/styles/default/textarea.json failed, reason: write EPROTO C05ED3DE01000000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:ssl/record/ssl3_record.c:355:

    at ClientRequest.<anonymous> (file:///Users/anthony/.npm/_npx/a9116f6e444a1e9e/node_modules/node-fetch/src/index.js:108:11)
    at ClientRequest.emit (node:events:515:28)
    at TLSSocket.socketErrorListener (node:_http_client:495:9)
    at TLSSocket.emit (node:events:515:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
[...]

PR #578, opened 2023-12-31
feat: add support for `update utils` command

Skylli202:
Hi,

I tried to implement the feature discussed in this issue #298.

The original issue is that there is no way to update the `utils.ts` file. I added a tiny bit of logic to enhance the `update` command by recognizing the special keyword `utils` as @huntabyte suggested in his [comment](https://github.com/huntabyte/shadcn-svelte/issues/298#issuecomment-1722601595).

If the CLI is called with only the argument `utils` then it stops after updating the `utils.ts` file. Otherwise, it continues as normal.

I'm not too much familiar with this code base, so please, correct me if:
- The utils file is actually in the registry, but from my exploration it is not.
- Tell me if it is normal for the `config.resolvedPaths.utils` value to not contain the file extension, like if it was a folder?

> From a previous contribution, I think I need to run a release command for semversionning. Please tell me what to do. I could not find any information on that matter in the `CONTRIBUTING.md`.

- [ ] Format & lint the code with `pnpm format` and `pnpm lint`

> `pnpm lint does not work in the folder `/package/cli`, I believe it should be replace by `pnpm format:check` which has been run and passed successfully.

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -17,6 +17,7 @@ import {
 	getRegistryIndex,
 	resolveTree
 } from "../utils/registry";
+import { UTILS } from "../utils/templates";
 import { transformImport } from "../utils/transformer";
 
 const updateOptionsSchema = z.object({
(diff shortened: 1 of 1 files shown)


Creative-Difficulty:
🎉 Awesome, thanks so much!

AdrianGonz97:
thank you!

PR #614, opened 2024-01-06
fix: pagination overflowing

### LINKED ISSUE #600: Pagination component overflowing its container on mobile devices
### Change Type

Correction

### Proposed Changes

As you can see in the screenshot, on mobile devices the pagination component overflows the container:
Overflowing pagination

I marked this as a docs issue since a user of this component can easily change the number of pages shown on mobile screens as they desire.

HubbeDev:
This pr attempts to fix #600. Im not 100% sure this is the right way to go. 

Feel free to come with feedback if their is a better approach. 

### The fix includes

Hide every page item except current page item and ellipsis.
I also hide the "Previous" and "Next" labels on small screen to make it more tidy. 


huntabyte:
Hey @HubbeDev, thanks for taking this on! I don't think we should change the `<Pagination.XX>` components. We should just update the `pagination-demo` files to ensure it is responsive.

huntabyte (changes requested):
Let's just update the demos to be responsive rather than the individual components. Those can remain as they are in the `main` branch!

huntabyte:
Actually what we could do is update the pagination next button and previous buttons to keep the existing API the same, but move it into a slot fallback, so something like this:

```svelte
<PaginationPrimitive.PrevButton asChild let:builder>
	<Button
		variant="ghost"
		class={cn("gap-1 pl-2.5", className)}
		builders={[builder]}
		on:click
		{...$$restProps}
	>
		<slot>
			<ChevronLeft class="h-4 w-4" />
			<span class="hidden sm:block">Previous</span>
		</slot>
	</Button>
</PaginationPrimitive.PrevButton>
```

This way, in our updated `pagination-demo` components, we could do something like:

```svelte
<Pagination.Item>
			<Pagination.PrevButton>
				<ChevronLeft class="h-4 w-4" />
				<span class="hidden sm:block">Previous</span>
			</Pagination.PrevButton>
</Pagination.Item>
```

And the same would go for the next button of course.

HubbeDev:
@huntabyte I have moved the logic outside the components now with a slot instead. 
I didn't know that svelte-legos look awesome. But I'm not 100% sure if this is the best way. The code works and updates the pagination but only on load. Not resize. So if the page is loaded on a desktop the pagination will still overflow when resizing the window.

huntabyte:
Oh, this is actually a bug with `bits-ui`, the options aren't being updated with the new props. Let me push a patch fix to bits and this should resolve itself with a version update.

huntabyte (approved):

huntabyte:
Just pushed a commit that fixes this issue! Thanks a ton @HubbeDev, you're greatly appreciated!

PR #661, opened 2024-01-17
docs(forms): :memo: add options to forms

LBF38:
This is a small PR for updating the forms documentation. It also fixes the form's data loading by awaiting all `superValidate` functions in `+page.server.ts` files.

Finally, the examples' forms now display sonner's toasts to show how to handle the form's results in the UI.
It will render different toasts depending on the form's result.

Enjoy ! 🎉

### DIFF
--- a/apps/www/src/content/components/form.md
+++ b/apps/www/src/content/components/form.md
@@ -4,7 +4,7 @@ description: Building forms with Formsnap, Superforms, & Zod.
 ---
 
 <script>
-	import { Steps, ComponentPreview, FormPreview } from '@/components/docs';
+ import { Steps, ComponentPreview, FormPreview } from '@/components/docs';
 </script>
 
 Forms are tricky. They are one of the most common things you'll build in a web application, but also one of the most complex.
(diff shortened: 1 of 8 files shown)


huntabyte (approved):
Incredible, thanks for this!

simond07:
Hey, great example!
Where can I view this? It seems that the [docs](https://www.shadcn-svelte.com/docs/components/) are not updated with this. Neither the [examples](https://www.shadcn-svelte.com/examples/forms).
Thanks in advance

LBF38:
> Hey, great example!
> Where can I view this? It seems that the [docs](https://www.shadcn-svelte.com/docs/components/) are not updated with this. Neither the [examples](https://www.shadcn-svelte.com/examples/forms).
> Thanks in advance

Here is the form example using sonner's toast : https://www.shadcn-svelte.com/docs/components/form

simond07:
The current example does not show the toast.loading(Submitting...) and the code there either. I was searching for an example with server and client side validation and showing a toast for the loading, the error and the success state. I found this here but neither the Form.Root element nor the formsnap 'option' are no longer available. I would be interested in a 'full' solution. I know that svelte superform has the [delayed](https://superforms.rocks/concepts/timers) property. Ideally that would be used to create the toast after a short delay if the server call was not successful. I think that would be a great example for the docs.

LBF38:
You can always test it on your side. 
And if you feel like this should be an example in the docs, you might consider making a PR to the project !

PR #702, opened 2024-01-25
fix: menu `SubContent` with min width

### LINKED ISSUE #700: Menubar SubContent not taking the width it needs
### Describe the bug

Shadcn-Svelte music example menu (File -> Library):
Screen 1

Shadcn-React music example menu (File -> Library):
Screen 2

### Reproduction

https://www.shadcn-svelte.com/examples/music
https://ui.shadcn.com/examples/music

### Logs

_No response_

### System Info

```bash
MacOS Chrome
```

### Severity

annoyance

HubbeDev:
This should fix #700. When I was looking at [shadcn-ui](https://ui.shadcn.com/docs/components/menubar) I noticed that they are wrapping the subcontent with 2 divs instead of one as we do here. But I do not see why it's necessary. 

This pr fixes the issue without making any bigger adjustments. Let me know if you disagree @huntabyte

### DIFF
--- a/apps/www/src/lib/registry/default/ui/menubar/menubar-sub-content.svelte
+++ b/apps/www/src/lib/registry/default/ui/menubar/menubar-sub-content.svelte
@@ -15,7 +15,7 @@
 	{transition}
 	{transitionConfig}
 	class={cn(
-		"z-50 min-w-[9rem] rounded-md border bg-popover p-1 text-popover-foreground focus:outline-none",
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
Thank you!

Scooter1337:
But if a list is smaller than 9/8rem, it will end up being smaller than that, right? Shouldn't we use calc here?

PR #741, opened 2024-02-04
fix: grey out Menubar.Sub when disabled

SkyfallWasTaken:
Small PR that fixes the Menubar.Sub style when it's disabled.
Here's how it looks like now:
image


huntabyte (approved):
Thank you!

PR #742, opened 2024-02-04
fix: pnpx -> pnpm dlx in copy button

sadfun:
Because pnpx is [deprecated](https://github.com/pnpm/pnpm/pull/3652) and doesn't work in v7+ versions on pnpm.


huntabyte (approved):
Thanks!

PR #797, opened 2024-02-22
chore: add event forwarding & types for default/ui/switch

flatpickles:
# Add event forwarding & types for default/ui/switch

PR #220 added these events to `new-york/ui/switch.svelte`, but not into the `default` version of this component. This small change closes that gap, and also adds `type $$Events` for further parity with the New York component.


huntabyte (approved):
Thank you! Apologies on the oversight!

PR #811, opened 2024-02-24
docs: add Mail example

shyakadavis:
### Description:

Add the Mail example

Fixes:

If you feel like you want to take over, @huntabyte, please go ahead. Otherwise, I might get back to this in a few.

### Notes:

- Since you mentioned that there would be no need to introduce a new dependency(s), i.e. [Svelte-Time](https://github.com/metonym/svelte-time)/`Day.js`, I temporarily commented out instances where I used it the first time, in hopes that `@internationalized/date` could easily find its way there.

### To-do/To-fix:
- [x] Select of the `Account Switcher` is giving me a challenge. e.g. Once collapsed, it should remain in its full width, and other styling to match the original...
- [ ] Use the `Scroll Area` component once implemented
- [x] ~~Use `@internationalized/date` to format the relative timestamps.~~
So, the package doesn't support relative timestamps and recommends the [native relative object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat), which is pretty limiting I think - full discussion [here](https://github.com/adobe/react-spectrum/discussions/2775). Used that util function in that discussion, but couldn't take it far.
- [x] Tooltips are still "jumping" around when interacting with the Calendar and nearby tooltip triggers.
Though I haven't tested, I believe this https://github.com/melt-ui/melt-ui/pull/1006 might provide a fix. IDK, just hoping.
- [x] I don't know which events are involved, but when re-sizing panes, text is being selected as you move over.
- [x] Cookies are failing on the next read/load and breaking everything with this error:
```
SyntaxError: Unexpected non-whitespace character after JSON at position 1
```
I tried laying the groundwork, but couldn't take it far. This file has the bug description: `/apps/www/src/routes/examples/mail/mail.svelte`
- [x] Clean-up (folder/file structure)


huntabyte:
Okay, I think everything except the scroll area is good. Check it out and let me know if I missed anything @shyakadavis

huntabyte:
I'm comfortable with where it's at right now. If you or anyone else catch anything odd, feel free to PR it!

anatolzak:
> Tooltips are still "jumping" around when interacting with the Calendar and nearby tooltip triggers.
Though I haven't tested, I believe this https://github.com/melt-ui/melt-ui/pull/1006 might provide a fix. IDK, just hoping.

@shyakadavis can you give an example?

shyakadavis:
Hi, @anatolzak 

Seems Hunter fixed it.

From what I read, it turns out that the `group` prop, when set, can only ever allow one tooltip to be visible at a time. (Something I totally skipped from the docs - https://www.bits-ui.com/docs/components/tooltip#tooltiproot)

So yeah, that was a TIL for me. 😅 

Btw, @anatolzak love your work. I find myself always reading your P.R's as much as do Hunter's. 😅

anatolzak:
@shyakadavis Glad to hear it was fixed! 

> Btw, @anatolzak love your work. I find myself always reading your P.R's as much as do Hunter's. 😅
I really appreciate that! :)

I am actually self-taught and have started looking for a job recently, which is seeming a lot more difficult that I expected. I created a portfolio https://anatolz.com showing some of what I have worked on from pure passion and interest during the last 3 years. Would love any feedback :)

shyakadavis:
> started looking for a job recently

Would you look at that, me too, @anatolzak! And I agree, it's brutal out here 😅 I've been looking for a while now, but nothing; but I hope we both land something soon. 🙂 

I've actually been meaning to reach out to you, but couldn't find any contact link. Do you mind opening your Twitter DM's to other people; currently you have them set to never receive messages from strangers (It's a default thing), or better yet, you could join Hunter's Discord 🙂

PR #817, opened 2024-02-25
feat: allow for default select icon to be hidden

MentalGear:
I'd like to propose adding an option to switch off the default `select icon` in the `Select Input`.
This would allow using custom icons with ease by providing them right in the slot, or none at all if the dev prefers.

Example:
Screenshot 2024-02-25 at 11 23 13


MentalGear:
Just noticed that the element received an update before my code adjustment - let me know if you agree with the general idea and we can see to update it to the new version. Cheers !

huntabyte:
Hey there, I'm cool with adding this prop.

Let's make the prop `withIndicator` defaulting to `true`. These changes should apply to both the `new-york` and `default` styles.

PR #823, opened 2024-02-26
fix(docs): mail example github code url

anatolzak:
the github url for the mail example in the docs leads to a 404


AdrianGonz97:
thank you!

PR #830, opened 2024-02-27
fix: add `for` to labels in default/example/dialog-demo.svelte

braden-w:
Added the html `for` attribute in labels for best practice. For some reason, this was featured in [New York](https://github.com/huntabyte/shadcn-svelte/blob/main/apps/www/src/lib/registry/new-york/example/dialog-demo.svelte) style but not default.


AdrianGonz97:
thank you!

PR #840, opened 2024-03-01
docs: fix a couple of words in the data table docs

Manuel-DaSilva:
Fixes two incorrect words in some sections of the data table. The phrase "to customize the pagination behavior" was repeated in the filtering and selecting steps.


huntabyte (approved):
Thank you!

PR #843, opened 2024-03-02
feat: add overflow-auto, remove sm:text-center to sheet

skelawsky:
This component would be much better out of the box if it had 'overflow-auto' set. Additionally, `text-center` on smaller screens doesn't look very good and it spoils the look of the forms by making them look worse. 

Before:
image
After:
image


huntabyte:
If this is not the behavior of the original https://ui.shadcn.com then this will not be added as a default here.

huntabyte:
Going to close this as the goal is to remain aligned with the original styles. If this is changed there, then we will update it here!

PR #869, opened 2024-03-07
fix: Edit import of "$lib/utils" in new component "ScrollArea" for consistency across the repo

ekpah:
Make a small edit in the import of a new component to stay consistent and prevent potential conflicts, if only "utils.ts" and not "utils.js" exists


huntabyte (commented):
This is actually an issue we need to address with the transformation. It should be `$lib/utils.js` across the entire project to support both bundlers and Node.

huntabyte:
The extension of the import should be what it is compiled to, which is `utils.js` not `utils.ts`. See https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries

PR #879, opened 2024-03-09
chore: Prerender docs experiments

AdrianGonz97:
- Every page under the `/docs/**` route is now prerendered
- Fixed the worker manifest such that static assets no longer triggering edge functions
- Moved `__registry__` inside of `src` so we're not fumbling around with vite's FS access (also adjusted the types for the preview component)


huntabyte (approved):
🫡

PR #890, opened 2024-03-11
fix escape sequence in `.eslintrc` example

### LINKED ISSUE #888: Installation instructions for .eslintrc should escape backslashes
### Change Type

Correction

### Proposed Changes

Copy-pasting the .eslintrc from the [docs](https://www.shadcn-svelte.com/docs/installation#eslint-configuration) leads to parsing errors via eslint/js-yaml.

```
YAMLException: Cannot read config file: ./src/lib/components/ui/.eslintrc
Error: unknown escape sequence (7:33)

 4 |       "warn",
 5 |       {
 6 |         "argsIgnorePattern": "^_",
 7 |         "varsIgnorePattern": "^\$\$(Props|Events|Slots|Generic)$"
-------------------------------------^
 8 |       }
 9 |     ]
```

They seem to be escaped in the markdown sources already. The straightforward solution would be double-escaping them [here](https://github.com/huntabyte/shadcn-svelte/blob/f391ad76dd21f5ccd7b85ea57505f291ce64a008/apps/www/src/content/installation/index.md?plain=1#L104), ~but I didn't yet verify if that works~.

psytrx:
fixes #888.

The double escape sequence does the trick, tested locally.

### DIFF
--- a/apps/www/src/content/installation/index.md
+++ b/apps/www/src/content/installation/index.md
@@ -101,7 +101,7 @@ One option is to add a `.eslintrc` file in the directory where you define your c
       "warn",
       {
         "argsIgnorePattern": "^_",
-        "varsIgnorePattern": "^\\$\\$(Props|Events|Slots|Generic)$"
+        "varsIgnorePattern": "^\\\\$\\\\$(Props|Events|Slots|Generic)$"
       }
     ]
   }


huntabyte (approved):
Thanks!

PR #896, opened 2024-03-13
docs: fix wrong line highlighting in form example

olegpolin:
In step 5 of the [form example](https://www.shadcn-svelte.com/docs/components/form), the highlighting of what lines need to be added is off. 

The last semicolon isn't highlighted, which could cause syntax errors for anyone following along with the example.

Shifting the highlighting by 1 line fixes this.


huntabyte (approved):
Thank you!

PR #910, opened 2024-03-15
fix: styles for `Select.Trigger`

JabSYsEmb:
Fix an issue in "/themes" page - report issue component:
Screenshot 2024-03-15 at 21 19 03
"line-clamp-1" is setting the display to "-webkit-box" which overwrite the "flex" class therefore I have remove "line-clamp-1" from the classes and add "overflow-hidden" instead.

Closes: #1020

### DIFF
--- a/apps/www/src/lib/registry/default/example/cards/report-issue.svelte
+++ b/apps/www/src/lib/registry/default/example/cards/report-issue.svelte
@@ -79,7 +79,7 @@
 				<Select.Root selected={securityLevels[1]}>
 					<Select.Trigger
 						id="security-level-{id}"
-						class="w-full overflow-hidden truncate"
+						class="w-full truncate"
 					>
 						<Select.Value placeholder="Select level" />
 					</Select.Trigger>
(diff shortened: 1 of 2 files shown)


huntabyte (approved):

huntabyte:
Thank you! @JabSYsEmb

Scooter1337:
2 things:
- Truncate already sets overflow to hidden, 'overflow-hidden' is not doing anything.
- ui.shadcn.com/themes uses line-clamp-1 too, do you have the same issue there?

JabSYsEmb:
@Scooter1337 
image

I wasn't aware that truncate sets the overflow to hidden, but surprisingly ui.shadcn.com/themes looks fine to me even though they merge "flex" + "line-camp-1"  together.  

I just noticed that ui.shadcn.com uses an old version of tailwind-merge:
```js
{
		"tailwind-merge": "^1.12.0",
}
```
whereas shadcn-svelte uses:
```js
{
		"tailwind-merge": "^2.2.1",
}
```

if you update the tailwind-merge dependency you'll face the same issue as here
Screenshot 2024-03-18 at 11 24 38

huntabyte:
What's the verdict here?

JabSYsEmb:
The "overflow-hidden" class is redundant as "truncate" also sets the overflow to "hidden".

Scooter1337:
@JabSYsEmb could you lint, format and push so the checks are all green? 😃

PR #921, opened 2024-03-19
Fix: ui fix for labels of sidebarNavItem

JabSYsEmb:
The labels of sidebarNavItem were erroneously displaying underlines despite having the no-underline class applied.

The no-underline class alone was not affecting the span element's style as expected. This is because the text-decoration property was being inherited from the parent <a /> element. By setting the display property of the label to "inline-block", we prevent it from inheriting the underline style when the anchor is hovered over.

before
Screenshot 2024-03-19 at 22 04 56

after
image

### DIFF
--- a/apps/www/src/lib/components/docs/nav/docs-sidebar-nav-items.svelte
+++ b/apps/www/src/lib/components/docs/nav/docs-sidebar-nav-items.svelte
@@ -13,7 +13,7 @@
 				<a
 					href={item.href}
 					class={cn(
-						"group flex w-full items-center rounded-md border border-transparent px-2 py-1 hover:underline",
+						"w-full rounded-md border border-transparent px-2 py-1 hover:underline",
 						item.disabled && "cursor-not-allowed opacity-60",
 						$page.url.pathname === item.href
 							? "font-medium text-foreground"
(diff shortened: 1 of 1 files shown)


huntabyte:
Hey there, the underline seems to be spanning beyond the word by a character!

JabSYsEmb:
a whitespace is added between the label and the text of the anchor element but I am not sure why it's added let me try to fix that also
image

JabSYsEmb:
hey,

There is two options to get rid of the whitespace after the text which are the following:
- wrap the `{item.title}` with a span element and set its margin right to -4px  as here:

```js
// ...
>
					<span class="-me-1">{item.title}</span>
					{#if item.label}
// ....
```
- to remove the space between html elements as follow:
```js
// ....
				>
					{item.title}{#if item.label}
						<span
							class="ml-2 inline-block self-center rounded-md bg-[#adfa1d] px-1.5 py-0.5 text-xs leading-none text-[#000000]"
						>
// ....
```
I quote from a website:
" A common problem when using inline-block elements is that whitespace in HTML becomes visual space in the site design (similar to having space between words). " [resource](https://university.webflow.com/lesson/remove-whitespace-between-inline-block-elements?topics=layout-design)

Which solutions do you think are more suitable in this situation?

huntabyte:
The space is fine, we just don't want it underlined. If we wrap the title in a `span` and apply the `underline` class to that it should work right?

huntabyte (approved):
Thank you!

AdrianGonz97:
Closing as this is fixed in `next` and we won't be adding anything new to the `main` branch.

Issue #883, opened 2024-03-09
Ring variable in app.pcss file is defined as a function


jesi-rgb:
### Describe the bug

The docs very conveniently explain that colors in the pcss file for the themes must be defined without the color space function.

The automatic installer, however, does include the --ring variable as a function in the `.dark` class selector:
```diff
  .dark {
    --background: 192.2 4% 10.9%;
    --foreground: 210 40% 98%;

    --muted: 217.2 4% 17.5%;
    --muted-foreground: 215 4% 65.1%;

    --popover: 192.2 4% 4.9%;
    --popover-foreground: 192.2 4% 98%;

    --card: 192.2 84% 4.9%;
    --card-foreground: 210 40% 98%;

    --border: 192.2 4% 17.5%;
    --input: 192.2 4% 17.5%;

    --primary: 192.2 4% 21%;
    --primary-foreground: 192.2 4% 90.2%;

    --secondary: 217.2 32.6% 17.5%;
    --secondary-foreground: 210 40% 98%;

    --accent: 192.2 9% 17.5%;
    --accent-foreground: 192.2 4% 98%;

    --destructive: 0 62.8% 30.6%;
    --destructive-foreground: 210 40% 98%;

+   --ring: hsl(212.7, 26.8%, 83.9);
  }
}
```

the bug can easily be solved removing the function and parentheses, but it was confusing since none of the ring behaviours I expected (focusing on textarea, for example) worked.

Very small but important!

### Reproduction

Install the latest svelte-shadcn version as usual

### System Info

```bash
System:
    OS: macOS 14.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.71 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.5.0 - /opt/homebrew/bin/node
    Yarn: 4.1.0 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - /opt/homebrew/bin/npm
    pnpm: 8.6.12 - ~/Library/pnpm/pnpm
    bun: 1.0.0 - ~/.bun/bin/bun
  Browsers:
    Chrome: 119.0.6045.159
    Safari: 17.1
  npmPackages:
    @sveltejs/kit: ^2.0.0 => 2.5.3 
    bits-ui: ^0.19.5 => 0.19.5 
    mode-watcher: ^0.2.2 => 0.2.2 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

[...]

PR #922, opened 2024-03-20
fix: remove HLS from slate ring variable

HubbeDev:
Fix for [#883](https://github.com/huntabyte/shadcn-svelte/issues/883)


huntabyte (approved):
Thanks @HubbeDev !

PR #925, opened 2024-03-20
docs: Update tasks example to be 1:1 with original

Scooter1337:

Before:
ss1

After:
ss2

### DIFF
--- a/apps/www/src/routes/examples/tasks/(components)/data-table-column-header.svelte
+++ b/apps/www/src/routes/examples/tasks/(components)/data-table-column-header.svelte
@@ -1,10 +1,13 @@
 <script lang="ts">
+	import EyeNone from "svelte-radix/EyeNone.svelte";
 	import ArrowDown from "svelte-radix/ArrowDown.svelte";
 	import ArrowUp from "svelte-radix/ArrowUp.svelte";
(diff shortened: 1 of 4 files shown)


huntabyte:
Hey there, thanks for taking the time to knock this out! I think we should update it to align with the original

Scooter1337:
> Hey there, thanks for taking the time to knock this out! I think we should update it to align with the original

Excuse me, I must have hallucinated the text. Fixed it, and made it 1:1 with the original.

Scooter1337:
ss3
Columns now hidable via header, also added icons to Asc and Desc, matching the original.

Scooter1337:
@huntabyte Could you look at my implementation of the 'Hide column' header functionality? I might be overcomplicating it.:

- now passing tableModel and cell.id to DataTableColumnHeader
- DataTableColumnHeader uses tableModel and cell.id to add the cell.id to hiddenCollumnIds
- DataTableViewOptions now watches for changes in tableModel's hiddenCollumnIds

Scooter1337:
Found out that 'Desc' always toggled 'Ascending' if props.sort.order was undefined. As there is no 'set' but only toggle, we now toggle twice if 'Desc' and props.sort.order === undefined. (undefined -> 'Asc' -> 'Desc')

PR #950, opened 2024-03-25
fix: make copy Button` visible in hex to channel inputs

carljohan:
Copy buttons for HSL and RGB inputs was using the same color as background, making them invisible on https://www.shadcn-svelte.com/docs/theming.  

- updated `CopyButton` color and hover color for these instances
- fix position of the CopyButtons
- moved `CopyButton` below input to focus on input before focus on copy buttons when navigating using keyboard
- remove tabindex from SVGs, which was causing double focus, both for button and SVGs when using navigating using keyboard. 

### Before
Screenshot-2024-03-25_11 25

### After
Screenshot-2024-03-25_11 20 2

### DIFF
--- a/apps/www/src/lib/components/docs/copy-button.svelte
+++ b/apps/www/src/lib/components/docs/copy-button.svelte
@@ -76,9 +76,9 @@
 			>
 				<span class="sr-only">Copy</span>
 				{#if copied}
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
Thank you!

PR #905, opened 2024-03-14
chore: Remove `@sveltejs/enhanced-img`

AdrianGonz97:
While `@sveltejs/enhanced-img` is a great plugin for optimizing images for SK, it unfortunately slows down our build pipeline by a significant amount. Previously, it would take a little under 2 minutes to complete a fresh build, whereas now it takes about 8 minutes. 

Currently, there's no available plugin/config option to cache the generated images, meaning that all images need to be processed (twice!) on _every_ build. Our little GH runner just can't keep up. 

We'll reevaluate once the ability to cache is implemented.

### Before submitting the PR, please make sure you do the following

- [ ] If your PR isn't addressing a small fix (like a typo), it references an issue where it is discussed ahead of time and assigned to you. In many cases, features are absent for a reason.
- [ ] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [ ] This message body should clearly illustrate what problems it solves.
- [ ] Format & lint the code with `pnpm format` and `pnpm lint`

PR #978, opened 2024-03-31
chore: Readd `@sveltejs/enhanced-img`

AdrianGonz97:
Follow-up for #905. Caching has now been implemented as of `v0.2.0`.

---

**Edit 1:** Looks like the images are cached to `node_modules/.cache/imagetools`, which is a bit of a problem due to how our pnpm cache is setup. We'll have to cache the images separately with `@actions/cache`.

---

**Edit 2:** After a painful amount of time debugging why the cache was never being used by `vite-imagetools`, I figured out that the problem is stemming from the generated image ID, which is a hash that's computed from 3 sources:
- the file's url
- the image's config object, stringified
- the `mtime` stat of the image file (which is acquired from `fs.statSync`)

https://github.com/JonasKruckenberg/imagetools/blob/8790098b61ab736c541ea3f15dff4026d9737e5d/packages/vite/src/utils.ts#L10-L21

That last value for the hash is the issue. It seems like the time calculated from that `Date` object is _relative_ to machine, which leads to a cache MISS when it's computed on the runner. In other words, the cache will only ever HIT on the machine that originally generated the cached images.

For now, I've added a local patch that replaces the `mtime` to use the `size` of the file instead.


AdrianGonz97:
Sticking with the pnpm patch for now. Will update if/when imagetools goes through

PR #989, opened 2024-04-04
fix: tasks example causing overflow on the page

Scooter1337:
## Overflow
Faced this issue in my own project, and remembered I copied it from this example.
~The original shadcn does not need this workaround.. can't figure out why...~

Before:
- Shrink the page horizontally so that the table starts overflow-scrolling by a big chunk (>100px), the sr-only text in the triple dots menu will start overflowing the page too.

https://github.com/huntabyte/shadcn-svelte/assets/11229056/802bb127-2e83-499f-9230-9d5ba0c06415

After:
~- By wrapping the sr-only text in a relative span, as explained by a _maintainer_ of tailwindcss [here](https://github.com/tailwindlabs/tailwindcss/discussions/12429), we can prevent the page from overflowing... nice!~

The original shadcn fixes this using relative on the table component, instead of per sr-only span.
Last commit fixes this.

https://github.com/huntabyte/shadcn-svelte/assets/11229056/313f9ae6-f7c0-4fc1-b130-3dd2704b4e89

## Faceted Filter
Before:
before

After:
after

TLDR:
- Fixed table component by adding relative, to fix overflowing bugs above
- Added icons and counts to the faceted filters

### DIFF
--- a/apps/www/src/lib/registry/default/ui/table/table.svelte
+++ b/apps/www/src/lib/registry/default/ui/table/table.svelte
@@ -8,7 +8,7 @@
 	export { className as class };
 </script>
 
-<div class="w-full overflow-auto">
+<div class="relative w-full overflow-auto">
 	<table class={cn("w-full caption-bottom text-sm", className)} {...$$restProps}>
 		<slot />
 	</table>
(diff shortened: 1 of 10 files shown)


huntabyte (approved):
Incredible. Thank you!

PR #1015, opened 2024-04-12
docs: add docs to skip confirmation prompt to `update` command

philip-weber:
I have added the option to skip the confirmation prompt to the `update` command to the cli docs. The option to the command is already present in the cli.

I have also written two description texts of options of the `add` command in lower case and thus aligned them with the other description texts. I hope it's okay that I have added this in this PR.


huntabyte (approved):
Thank you!

PR #1021, opened 2024-04-15
fix: block - default dashboard-06 table columns

reesebre82:
On the dashboard block, dashboard-06, some of the columns have the class table.-cell. I have went in and replaced them with table-cell.

Without this fix, the columns with table.-cell will not show at any screen width.


huntabyte (approved):

huntabyte:
Thank you!

PR #1023, opened 2024-04-15
feat: data-attribute-remover

Scooter1337:
Just an example, the data-x-block attributes are a bit annoying when copying the code.

Added function that uses regex to remove the data-x-block-name and description attributes and leaves a clean element.
```
<Card.Root data......" class="abc"> becomes <Card.Root class="abc">
<Card.Root data......"> becomes <Card.Root>
```
No leftover whitespace AFAIK.


huntabyte (approved):
Thank you!

PR #1026, opened 2024-04-17
fix: Self closing tags issue

TheOnlyTails:
Fixes the issue raised by Rich Harris here: https://github.com/sveltejs/svelte/issues/11052, wherein it is pointed out how self closing tags in HTML are a lie, and `<div />` is equivalent to an unclosed `<div>`.
I simply ran the self closing tags migration provided by `svelte-migrate` on the project.


huntabyte (approved):
Thanks for taking care of this!

Issue #1062, opened 2024-05-01
svelte-sonner slots


Carlos-err406:
### Describe the feature

the original svelte-sonner provides slots for modifying the default iconst of the toast
see [this section of the svelte-sonner readme](https://github.com/wobsoriano/svelte-sonner#changing-icon)

my request is to add those slots to the installed component to make it more extendable, and so users dont have to modify the /ui/sonner.svelte component just to add them

Carlos-err406:
also i would be glad to work on this, and if needed update the docs

huntabyte:
That would be great @Carlos-err406 !

Carlos-err406:
@huntabyte Do you think it's sane to copy and paste the Icon and Loader components from [svelte-sonner](https://github.com/wobsoriano/svelte-sonner/tree/main/src/lib), to use as default values of the slots, or is it better to get a quick PR on svelte-sonner to export the Icon component (and Loader) to the index, i dont know their stand on that though?

or maybe is preferable to just use lucide-svelte icons and forget about the defaults of svelte-sonner?

Carlos-err406:
i tried the lucide icons approach, i think it came out pretty nice

oliverdowling:
I've done this in a simpler way, by just forwarding the slots if they have been specified by using `{#if $$slots...` which doesn't require copy+paste of the other components or add an import.

```
{#if $$slots["loading-icon"]}<slot slot="loading-icon" name="loading-icon" />{/if}
{#if $$slots["success-icon"]}<slot slot="success-icon" name="success-icon" />{/if}
{#if $$slots["error-icon"]}<slot slot="error-icon" name="error-icon" />{/if}
{#if $$slots["warning-icon"]}<slot slot="warning-icon" name="warning-icon" />{/if}
{#if $$slots["info-icon"]}<slot slot="info-icon" name="info-icon" />{/if}
```

Happy to make a PR if you would like.

huntabyte:
With snippets coming up in `svelte-sonner@next` (not yet published) this issue will no longer be relevant.

PR #1066, opened 2024-05-03
feat: sonner slots for modifying icons

Carlos-err406:
addresses #1062 

- updated svelte-sonner version since the `locked` one does not have the slots feature


huntabyte:
Hey there! Thanks for taking the time to add this.

I'm cool with exposing the slots, but not sure how I feel about setting those default icons as the fallbacks.

I will experiment a bit with it to see how it feels!

Carlos-err406:
of course! thank you!

Carlos-err406:
@huntabyte maybe if this gets merged on svelte-sonner we dont have to use lucide-icons as fallbacks.

https://github.com/wobsoriano/svelte-sonner/pull/96

Carlos-err406:
now using same fallbacks as [svelte-sonner](https://github.com/wobsoriano/svelte-sonner) since now Icon and Loader are exported

huntabyte:
Use snippets in next.shadcn / svelte-sonner 1.0+

PR #1078, opened 2024-05-07
docs: fix svelte-add command syntax

yeeterlol:
addresses svelte-add changing command syntax and updates the docs to fix this


AdrianGonz97:
thank you!

PR #1082, opened 2024-05-08
fix: default textarea event types

joe-ervin05:
The default textarea component was missing the custom event types. This fix makes the components' event typing work the same way as other components do.

The new york variant is already typed correctly so only the default needs to be changed.


huntabyte (approved):
Thank you!

PR #1103, opened 2024-05-20
fix: superForm validation errors in examples that use froms

ttytm:
Currently, examples that use forms with validation and a destructured export of data run into an internal error, namely:

```
SuperFormError: No form data sent to superForm. Make sure the output from superValidate is used (usually data.form) and that it's not null or undefined. Alternatively, an object with default valu
```

The PR updates the export to match the one used in the example of the Form component itself and other working examples that use forms.


huntabyte:
It's exported like that for a reason to accommodate the documentation at the moment.

Issue #1037, opened 2024-04-20
Inconsistent Style Application for Placeholder in Select Trigger Component


armchair-traveller:
### Describe the bug

The Select Trigger component currently uses the `placeholder:text-muted-foreground` style which is not being applied correctly. As per defaults, it should look gray when applied. The issue stems from the `:placeholder` pseudo-element not being associated with buttons but rather with specific input controls.

### Proposed Fix
To resolve this issue, it is suggested to modify the CSS selector to `[&>span[data-placeholder]]:text-muted-foreground`. This approach is consistent with the existing `[&>span]:line-clamp-1` style. It is important to note that in our library, the `data-placeholder` attribute is applied to the value component rather than the trigger, which differs from the original library's implementation.

### Additional Information
This styling inconsistency is also present in the select component maintained by @shadcn. The fix for his component may differ slightly due to the implementation variations. I encourage any contributor to verify if a similar issue exists in his repository and report it if it isn't.

### Reproduction

Simply see the example component in [our Select docs](https://www.shadcn-svelte.com/docs/components/select) or @shadcn's [Select docs](https://ui.shadcn.com/docs/components/select) and observe that placeholder styles aren't being applied.

### System Info

```bash
N/A
```

### Severity

annoyance

huntabyte:
Closed by #1125

PR #1125, opened 2024-06-07
chore: select component improvements

herkulano:
### Select
- Fix placeholder https://github.com/huntabyte/shadcn-svelte/issues/1037
- Fix aria-invalid border
- Fix focus. It shouldn’t have focus on press. It should have focus upon exit/selection, but it seems melt-ui is not triggering it: https://melt-ui.com/docs/builders/select

### Colors
- Focus color was too dim on dark mode


huntabyte (approved):
thank you!

PR #1129, opened 2024-06-09
docs:Removed invalid destructuring that was causing me confusion

CodeDoes:
I think there was a change in the code that made this line invalid. Am I wrong?


shyakadavis:
Hmm... pretty sure the destructuring is correct.

`enhance` is used down below on `use:enhance` & `formData` store is used to access stuff like errors, bindings...

huntabyte:
Yep, destructuring is correct here.

Issue #1126, opened 2024-06-07
Reopening with reproduction: Data table becomes empty when email column removed #1117


brucegl:
### Describe the bug

HI,

I followed the guide in the Data Table section and when I hide the email column then all the rows are removed, and the row selection checkbox is checked.

### Reproduction

I reproduced this code on Stackblitz [here](https://stackblitz.com/edit/github-ihthdo-g2rp5k?file=src%2Froutes%2F%2Bpage.svelte) and I added the fix (i.e. adding the line, `includeHiddenColumns: true`, which is currently commented out)

If you uncomment the line, you can then hide the email column without removing all rows.

### System Info

```bash
Stackblitz
```

### Severity

annoyance

huntabyte:
So is this just an issue with the guide? If so, PRs welcome 😃

brucegl:
Ok, I'll do that now :-)

brucegl:
submitted PR https://github.com/huntabyte/shadcn-svelte/pull/1133 to close this

PR #1133, opened 2024-06-10
docs: fix empty data table when the email column is hidden

brucegl:
As per #1126, this is to fix an issue in the guide code (in the Data Table section), whereby, when I hide the email column then all the rows are removed, and the row selection checkbox is checked.


huntabyte (approved):

huntabyte:
Thank you!

Issue #1059, opened 2024-04-30
Not prompt me to pick what package manager to use everytime im copying the command


tryggs:
### Describe the feature

The gesture is nice, but it gets annoying and I've mistakenly pasted the same thing in the command line multiple times because I thought that I had copied it but in reality, it still wants me to pick the package manager. How often does one switch what package manager you want to use?

Ask me once, remember, and have the option to change below the code block would be my suggestion.

Michael-Obele:
I'd love to fix this! In fact, I was just thinking about suggesting it myself. I think using cookies should definitely work for this, and I'm happy to make a pull request to implement it.

huntabyte:
All you! @Michael-Obele could probably use a svelte-persisted-store for this

Michael-Obele:
I've made a PR, and I'd love comments on the chosen implementation and design choice.

Michael-Obele:
I made a new PR: Implemented the suggestions using the [Stripes UI](https://docs.stripe.com/libraries?lang=node) and added svelte-persisted-store (missed it last time).

tryggs:
Very cool @Michael-Obele! 

This is basically exactly how I imagined it, although I had two thoughts:

1. Design: maybe that wasn't the proposed final design, but it sticks out a little bit, especially with the border, and I think it should be a bit more muted.

2. The command text: currently it's always showing the npm command, but wouldn't it be better if it was showing the command that is the users preferred package manager. So many times, even though there is a copy button, I find myself just selecting the text in the code block and copying, but maybe it's just me.

Those are just my two cents, but functionality wise this solves the problem I was facing when I first presented the idea.

So good job :)

Michael-Obele:
Thanks, @tryggs. The first point is very valid; I'll change the code to make the border a little more mutted. The second point is also an excellent idea, but I'm hesitant about implementing it because I still believe it falls outside the scope of the main issue. I'll be making some adjustments to include the first (and hopefully the second) suggestion.

Thank you for your time reviewing this.

tryggs:
That is very true @Michael-Obele

Michael-Obele:
@tryggs and @huntabyte I have made the recommended changes; please check it out and tell me what you think.
FYI: I didn't implement the second suggestion; it'll be best to open a different issue for it, and if it is approved, we can work towards implementing it.

tryggs:
[...]

PR #1181, opened 2024-07-12
feat: Persist user's package manager choice

Michael-Obele:
**Addressing issue #1059: Persistent Package Manager Selection**

This PR enhances user experience by introducing a persistent package manager selection feature. The user's chosen package manager is now saved and automatically used across pages within a session, eliminating the need for repeated prompts.

**Key Improvements:**

* **Persistent storage:** User preferences are stored using a writable store for easy retrieval.
* **UI enhancements:** The selected package manager is clearly displayed with an option to change it.
* **Improved user experience:** Reduces user friction by eliminating repetitive prompts.

**Changes:**

* Integrated `selectedCommand` writable store to manage user preferences.
* Updated `copy-button.svelte` to use the selected package manager.
* Enhanced UI to display and allow modification of the package manager.

**Note:** The selected package manager is currently reset on page reload.

## Showcase

[Screencast from 2024-07-12 05-01-54.webm](https://github.com/user-attachments/assets/c766cb5c-7057-4ed8-a44f-36fbf1981167)

### DIFF
--- a/sites/docs/src/content/components/sonner.md
+++ b/sites/docs/src/content/components/sonner.md
@@ -45,6 +45,8 @@ npx shadcn-svelte@latest add sonner
 	Add the Toaster component
 </Step>
 
+Note: Make sure you are adding the import from the path `"$lib/components/ui/sonner"` not `"svelte-sonner"`.
+
 ```svelte title="+layout.svelte" {2,5}
 <script lang="ts">
   import { Toaster } from "$lib/components/ui/sonner";


huntabyte:
Hey @Michael-Obele ! Thanks for taking this on! I'll give it a review tomorrow!

huntabyte:
Hey @Michael-Obele, the functionality is great. I'm not sure how I feel about having the package manager's name show up beside the icon like that, though. It seems a bit unintuitive. Perhaps we can set it in the `CMD+K` menu (similar to themes), where we have an option like "Set Default Package Manager" or "Clear Default Package Manager" for power users.

If they have it set, it just copies the command for their specific PM. Otherwise, it shows the dropdown for the non-power users.

tryggs:
Good job!

As @huntabyte said, the functionality is there but for me at least, I feel like its a bit unclear what is going on. Now the copy button still prompts the user to pick libary, and that the text (bun, npm, etc) copies the selected value.

How would a dropdown next too the copy button look? The value of the dropdown would be saved in local storage.

See [Stripes codeblock](https://docs.stripe.com/libraries?lang=node) in their docs.

huntabyte:
> Good job!
> 
> 
> 
> As @huntabyte said, the functionality is there but for me at least, I feel like its a bit unclear what is going on. Now the copy button still prompts the user to pick libary, and that the text (bun, npm, etc) copies the selected value.
> 
> 
> 
> How would a dropdown next too the copy button look? The value of the dropdown would be saved in local storage.
> 
> 
> 
> See [Stripes codeblock](https://docs.stripe.com/libraries?lang=node) in their docs.

Oh I quite like this idea as well!

Michael-Obele:
> Hey @Michael-Obele, the functionality is great. I'm not sure how I feel about having the package manager's name show up beside the icon like that, though. It seems a bit unintuitive. Perhaps we can set it in the `CMD+K` menu (similar to themes), where we have an option like "Set Default Package Manager" or "Clear Default Package Manager" for power users.
> 
> If they have it set, it just copies the command for their specific PM. Otherwise, it shows the dropdown for the non-power users. What do you think about that?



This would be nice, but I can see how people would easily get confused by the options of set and clear, I would suggest, however, that if we want to go with this design, we use a single click for the default (which would be whatever was last used) and a right click/context menu to select a new default. I use multiple package managers so I tend to switch often, so I hope to make the switch as easy as possible.

Michael-Obele:
> Hey @Michael-Obele ! Thanks for taking this on! I'll give it a review tomorrow! One thing I do see is that the bun lockfile will need to be removed before merge, as we use `pnpm` 😃

Got it, I'll do that. 😊

Michael-Obele:
> Good job!
> 
> As @huntabyte said, the functionality is there but for me at least, I feel like its a bit unclear what is going on. Now the copy button still prompts the user to pick libary, and that the text (bun, npm, etc) copies the selected value.
> 
> How would a dropdown next too the copy button look? The value of the dropdown would be saved in local storage.
> 
> See [Stripes codeblock](https://docs.stripe.com/libraries?lang=node) in their docs.

I think this design is the best because it is simple to use, requires less pasting from an unknown manager because you can always tell which manager you are on. The local stage would probably be required. 
If we're okay with the design, I can implement it, and we can see how it looks.

tryggs:
> > Good job!
> > As @huntabyte said, the functionality is there but for me at least, I feel like its a bit unclear what is going on. Now the copy button still prompts the user to pick libary, and that the text (bun, npm, etc) copies the selected value.
> > How would a dropdown next too the copy button look? The value of the dropdown would be saved in local storage.
> > See [Stripes codeblock](https://docs.stripe.com/libraries?lang=node) in their docs.
> 
> I think this design is the best because it is simple to use, requires less pasting from an unknown manager because you can always tell which manager you are on. The local stage would probably be required. If we're okay with the design, I can implement it, and we can see how it looks.

Me, personally, do think that it would be the best alternative, since it makes it very clear what you are copying, while always being one click away. Make sure that the content of the code block is the correct one for the selected value 😄

Michael-Obele:
> > > Good job!
> > > As @huntabyte said, the functionality is there but for me at least, I feel like its a bit unclear what is going on. Now the copy button still prompts the user to pick libary, and that the text (bun, npm, etc) copies the selected value.
> > > How would a dropdown next too the copy button look? The value of the dropdown would be saved in local storage.
> > > See [Stripes codeblock](https://docs.stripe.com/libraries?lang=node) in their docs.
> > 
> > 
> > I think this design is the best because it is simple to use, requires less pasting from an unknown manager because you can always tell which manager you are on. The local stage would probably be required. If we're okay with the design, I can implement it, and we can see how it looks.
> 
> Me, personally, do think that it would be the best alternative, since it makes it very clear what you are copying, while always being one click away. Make sure that the content of the code block is the correct one for the selected value 😄

Yes, I think it is great, but it might not be so simple to display the right text depending on the package manager. Essentially, the copy function simply modifies the necessary portions of the text (such as the npx to bunx) because the code is hardcoded in a markdown format. We should definitely ask @huntabyte. I could simply write a function to change the display, but I am not sure if that would be necessary.

tryggs:
How's it going?

I decided today that wanted to learn how to contribute to open-source projects (something I've never done before), and since I haven't seen this being implemented yet, I was thinking maybe I could give it a go.

Michael-Obele:
Hey @tryggs 
Though I have not heard back regarding the suggested implementation, I am still working on this.
I have already implemented the updated design, which resembles the [Stripes codeblock](https://docs.stripe.com/libraries?lang=node). If everything is okay, I will update the PR and see if this can be merged shortly. I would really like to see this issue fixed.

PR #1184, opened 2024-07-14
docs: make button size in site-header uniform

ttytm:
Current:
|shadcn-svelte|shadcn/ui|
|-|-|
|Screenshot_20240714_132306Screenshot_20240714_132257|Screenshot_20240714_132333Screenshot_20240714_132337|

The update makes the button sizes uniform and adapts them to the upstream style.
It does so by using the same tailwind classes that shadcn is using (ref. https://github.com/shadcn-ui/ui/blob/f170784f78b553641abbfe17c76a500a9563cf50/apps/www/components/site-header.tsx#L33). Personally, for squares, I would use the `size-` class, which combines `h-` and `w-`.

I did check the next branch to make sure the change that I'm not re-authoring a change which is already included.

<br>

---

As a justification, since the contribution is missing the first step of the contribution guidelines of opening an issue first:
 
The rationale of the guidelines is to save a contributor's time. In regards to the submitted changes, the most time-efficient approach for me was to open a descriptive PR directly. I'm okay if it's closed and not merged, as this would still save time for me. Please don't hesitate to close it if the changes are not appropriate. In the end, this might save time on your side as well. Apologies for not following the guidelines.

### DIFF
--- a/apps/www/src/lib/components/docs/mode-toggle.svelte
+++ b/apps/www/src/lib/components/docs/mode-toggle.svelte
@@ -8,7 +8,7 @@
 
 <DropdownMenu.Root>
 	<DropdownMenu.Trigger asChild let:builder>
-		<Button builders={[builder]} variant="ghost" class="w-9 px-0">
+		<Button builders={[builder]} variant="ghost" class="h-8 w-8 px-0">
 			<Sun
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
Thank you!

PR #1185, opened 2024-07-14
chore: update superseded package

ttytm:
The PR updates a superseded dependency that doesn't receive updates anymore because it was moved to a new package name.

Ref.: https://www.npmjs.com/package/svelte-local-storage-store

<br>

I'm adding a little justification, since the contribution is missing the first step of the contribution guidelines of opening an issue first:

- Based on the issue templates, I couldn't classify a category that would fit to open an issue prior to the PR. 
- The rationale of the guidelines is to save a contributor's time. The most time-efficient way I could think of for this specific change was to open the PR directly. I'm okay if it's closed and not merged, as this would have saved time for me. Please don't hesitate to close it if the changes are not appropriate.

### DIFF
--- a/apps/www/package.json
+++ b/apps/www/package.json
@@ -79,7 +79,7 @@
 		"paneforge": "^0.0.2",
 		"svelte-headless-table": "^0.18.2",
 		"svelte-legos": "^0.2.2",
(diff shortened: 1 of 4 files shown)


huntabyte (approved):
Thank you!

PR #1186, opened 2024-07-14
feat: Add flags for `init` command

### LINKED ISSUE #1183: Add choice flags for `init` command
### Describe the feature

Add flags for `init` to allow selection of style, base color,  and `-y` flag for prompts with sensible defaults. 

Why?: Nice to have if you are building automation tools such as templates that integrate the library

Proposed flags:
```
Usage: shadcn-svelte init [options]
 
initialize your project and install dependencies
 
Options:
  -c, --cwd <cwd>    the working directory. (default: the current directory)
  ---
  --ts                     use typescript
  -s, --style              the style ("default", "new york")
  --bc, --base-color       the base color for the components (colors...)
  --gc, ,--global-css      the global css file location
  --tc, --tailwind-config  the tailwind config location
  --ca, --component-alias  import alias for components
  --ua, --utils-alias      import alias for utils 
  ---
  -h, --help               display help for command
```

Happy to open a PR for this if its something you are open to!

ieedan:
# feat: Add flags for `init` command

- Fixes #1183

Adds new flags to allow for automated use of the CLI. This can be useful for anyone building templates around the library.

### How it works
User provides the flags for each option. If an invalid flag is provided it will throw an error (this is to prevent hanging of runs in automation tools). If a flag is not provided the user will be prompted the same way as before including retrys.

New flags
```
-ts, --typescript              use TypeScript
-s, --style <name>             the style (choices: "default", "new-york")
-bc, --base-color <name>       the base color for the components (choices: "slate", "gray", "zinc", "neutral", "stone")
-gc, --global-css <path>       path to the global css file
-tc, --tailwind-config <path>  path to the tailwind config file
-ca, --component-alias <path>  import alias for components
-ua, --utils-alias <path>      import alias for utils
```

New `--help` with the new flags
```
Usage: shadcn-svelte init [options]

initialize your project and install dependencies

Options:
  -c, --cwd <cwd>                the working directory. defaults to the current directory. (default: "E:\\github\\shadcn-svelte\\packages\\cli")
  -ts, --typescript              use TypeScript
  -s, --style <name>             the style (choices: "default", "new-york")
  -bc, --base-color <name>       the base color for the components (choices: "slate", "gray", "zinc", "neutral", "stone")
  -gc, --global-css <path>       path to the global css file
  -tc, --tailwind-config <path>  path to the tailwind config file
  -ca, --component-alias <path>  import alias for components
  -ua, --utils-alias <path>      import alias for utils
  -h, --help                     display help for command
```

### I had to make a few changes to make this work:
- Have to get the styles and base colors at the top level to be able to add them as choices. Because of this I had to change `getStyles` and `getBaseColors` into sync functions `getBaseColors` was already sync but `getStyles` needed to be changed to no longer fetch from the registry
- Had to ungroup the prompts so that I didn't have to do weird stuff with promises.

### Checklist

- [x] update cli docs
- [x] format and lint


ieedan:
This will require a bump in the CLI version not should it go to v0.10.8?

apps/www/src/content/cli.md
```diff
@@ -33,8 +33,15 @@ Usage: shadcn-svelte init [options]
 initialize your project and install dependencies
 
 Options:
-  -c, --cwd <cwd>  the working directory. (default: the current directory)
-  -h, --help       display help for command
+  -c, --cwd <cwd>                the working directory. defaults to the current directory. (default: "E:\\github\\shadcn-svelte\\packages\\cli")
```
huntabyte, line 36:
```suggestion
  -c, --cwd <cwd>                the working directory. defaults to the current directory.
```

huntabyte:
Thanks for taking the initiative to implement this! @AdrianGonz97 built/maintains the CLI so we'll wait for his review/comments before continuing forward!

ieedan:
Any update on this?

ieedan:
Yeah I can add that no problem. I have a few minutes now but if not in the next few minutes I can work on it this evening.

ieedan:
Hey so the validation for import aliases is a dependency of the TypeScript prompt? Is it okay if it goes after? 

```ts
const validateImportAlias = (alias: string) => {
  // `tsconfigName` is from the `Would you like to use TypeScript?` prompt.
  const tsconfig = cliConfig.getTSConfig(cwd, tsconfigName);
  // -- snip --
};
```

ieedan:
Maybe it would be better to just automatically detect tsconfig / jsconfig? I can't see a reason why we couldn't just do that?

ieedan:
We can detect with something like this:

```ts
export function detectLanguage(cwd: string): "tsconfig.json" | "jsconfig.json" | undefined {
	if (fs.existsSync(path.join(cwd, "tsconfig.json"))) {
		return "tsconfig.json";
	} else if (fs.existsSync(path.join(cwd, "tsconfig.json"))) {
		return "jsconfig.json";
	} else {
		return undefined;
	}
}
```

Ideally this would make the CLI more ergonomic for normal users as well.

AdrianGonz97:
> Hey so the validation for import aliases is a dependency of the TypeScript prompt? Is it okay if it goes after?
> 
> Edit: See below
> 
> ```ts
> const validateImportAlias = (alias: string) => {
>   // `tsconfigName` is from the `Would you like to use TypeScript?` prompt.
>   const tsconfig = cliConfig.getTSConfig(cwd, tsconfigName);
>   // -- snip --
> };
> ```

Sure, feel free to extract that out into it's own utility function and have the `cliCOnfig` and `tsconfigName` be passed in as arguments

> We can detect with something like this:
> 
> ```ts
> export function detectLanguage(cwd: string): "tsconfig.json" | "jsconfig.json" | undefined {
> 	if (fs.existsSync(path.join(cwd, "tsconfig.json"))) {
> 		return "tsconfig.json";
> 	} else if (fs.existsSync(path.join(cwd, "jsconfig.json"))) {
> 		return "jsconfig.json";
> 	} else {
> 		return undefined;
> 	}
> }
> ```
> 
> Ideally this would make the CLI more ergonomic for normal users as well.

This would not be ideal as the `tsconfig/jsconfig` can reside in parent directories. It would be best just to leave that job to `getTsconfig` from the `get-tsconfig` package

ieedan:
Would it be possible to detect the config using getTSConfig? It seems like it is locating it based on the directory containing the `package.json`.

Something like:
```ts
export function detectLanguage(cwd: string): "tsconfig.json" | "jsconfig.json" | undefined {
	if (cliConfig.getTSConfig(cwd, "tsconfig.json")) {
		return "tsconfig.json";
	} else if (cliConfig.getTSConfig(cwd, "jsconfig.json")) {
		return "jsconfig.json";
	} else {
		return undefined;
	}
}
```

ieedan:
Sorry for the spam I think this could work. This is exactly what the `cliConfig.getTSConfig` function is doing it just doesn't throw and retries once if it doesn't find it.

```ts
export function detectLanguage(cwd: string): "tsconfig.json" | "jsconfig.json" | undefined {
	const rootPath = path.resolve(cwd, "package.json");
	const tsConfig = getTSConfig(rootPath, "tsconfig.json");
	if (tsConfig != null) 
		return tsConfig;

	const jsConfig = getTSConfig(rootPath, "jsconfig.json");
	if (jsConfig != null) 
		return jsConfig;

	return undefined;
}
```

AdrianGonz97:
>Sorry for the spam I think this could work

No problem, I was just in the middle of writing a solution similar to that 🙂 I think that could work, go for it!

ieedan:
> > Sorry for the spam I think this could work
> 
> No problem, I was just in the middle of writing a solution similar to that 🙂 I think that could work (will have to tweak the return type to be the config type, but same idea), go for it!

Oops! Yeah I forgot to change that in my example.

ieedan:
We now auto detect TypeScript and that allows us to move the errors to the top when using flags.

Updated changeset and docs accordingly.

AdrianGonz97 (approved):
Perfect, this is fantastic! Thanks so much for your work on this!!

ieedan:
> Perfect, this is fantastic! Thanks so much for your work on this!!

No problem thanks for taking the time today!

PR #1192, opened 2024-07-18
fix: Some flags were being incorrectly assigned to config values

ieedan:
- Fixes an bug when using flags created by #1186

Some config values were being set to the tailwindConfig flag

Recreate with the following command
```
npm run start -- init -c testing --style new-york --base-color zinc --css src/app.css --tailwind-config tailwind.config.ts --components-alias '$lib/components' --utils-alias '$lib/utils.ts'
```

Gives this error:
```
[Config Error]: Invalid import alias found: ("utils": "tailwind.config.ts") in components.json.
   - Import aliases must use existing path aliases defined in your tsconfig.json (e.g. "utils": "$lib/utils").
   - See: https://www.shadcn-svelte.com/docs/installation/manual#configure-path-aliases.
```

I have renamed the flags and added a changeset for a patch.


ieedan:
@AdrianGonz97 if you have a chance can you take a look at this

AdrianGonz97:
Ah, completely my fault. Thanks for the fix!

AdrianGonz97:
FYI: it looks like github actions are super slow/at a standstill atm, so it might be a bit before the versioning PR is opened and a new release is cut. I'll keep an eye out for it as soon as it does 👍

ieedan:
> github actions are super slow/at a standstill atm

Impossible

ieedan:
We may need to restart the runners they seem to be frozen and GitHub is reporting normal now.

AdrianGonz97:
Yea, that workflow was completely bugged out. I couldn't even restart it, github was ignoring all requests to manually restart it. Instead, I had to make a new PR to retrigger the workflow. It should be good to go now

Issue #1149, opened 2024-06-17
[Separator] Vertical separator not rendering


TheGroobi:
### Describe the bug

Whenever the separator element has the orientation of "vertical" it sets a tailwind class of h-full which for whatever reason does not render the element at all (it has 1x0px  inside the dev tools).

What i found is that you can just remove the class completely from the SeparatorPrimitive.Root and it works fine, or add "min-h-full"

```svelte
class={cn(
	"shrink-0 bg-border",
	orientation === "horizontal" ? "h-[1px] w-full" : "min-h-full w-[1px]",
	className
)}
```

or 
```svelte
class={cn(
	"shrink-0 bg-border",
	orientation === "horizontal" ? "h-[1px] w-full" : "w-[1px]",
	className
)}
```

Both work as i tested on Brave and firefox.

### Reproduction

[StackBlitz](https://stackblitz.com/edit/github-fzhplc?file=src%2Froutes%2F%2Bpage.svelte)

### System Info

```bash
not neccesseray
```

### Severity

annoyance

Stadly:
I ran into the same problem. It happens when the parent container doesn't have a defined height. `min-h-full` seems like a good solution.

https://chatgpt.com/share/3e6600ef-fe9e-4260-9a8e-2d273c5a8975

huntabyte:
What @Stadly said!

Stadly:
Should I provide a PR for this, @huntabyte?

PR #1206, opened 2024-07-24
fix: Vertical separator height

Stadly:
Fixes height of vertical separator when parent container doesn't have a defined height. Closes https://github.com/huntabyte/shadcn-svelte/issues/1149.


huntabyte (approved):
Thank you!

PR #1211, opened 2024-07-24
feat: Make card header optional

Stadly:
This PR removes bottom padding from the card header instead of removing top padding from the card content. This way, the card content has consistent padding regardless of the presence of a header.

This is symmetrical to how it is done in the card footer.


Stadly:
This issue has also been reported in the main repo, but seems to not have been fixed: https://github.com/shadcn-ui/ui/issues/2533

huntabyte (approved):

huntabyte:
I think this is a good idea.

PR #1222, opened 2024-08-01
fix: dropdown-menu-shortcut.svelte to work in rtl layouts

jonsamwell:
Updated `dropdown-menu-shortcut.svelte` to use margin-inline-start (`ms-auto`) so it works in rtl layouts

Before fix:

image

With fix applied:

image


AdrianGonz97:
Closing as stale. However, we'll keep the related issue open.

PR #1235, opened 2024-08-13
cli changes

fernandolguevara:
- ~~`getPackageManager` try with user defined agent first~~
- log which package manager is executing dependencies installation imagen
- process only missing deps on `Installing package dependencies` step


AdrianGonz97:
Thank you, but I think this is doing a bit too much. I do like the idea of specifying which package manager is being used, but besides that, I'm not a fan of the other changes.

It may be easier to start a new PR (especially since we just changed how package managers are detected), so please feel free to open a new one with just the package manager change. thanks!!

PR #1243, opened 2024-08-17
docs: fix nav link overlapping issues

engageintellect:
My previous contribution (tailwind colors page) introduced a new link "colors" to the navbar. This causes the search bar to cover nav links on md-lg viewport widths.

Screenshot 2024-08-17 at 1 26 20 PM

You can also see in this photo that "shadcn-svelte" gets wrapped when squished. Also, the github link is being covered by the search bar.

I opted to remove the "GitHub" link, as it could be considered redundant, being that there is a github icon to the right of the search bar. 

I also mimicked the navbar behavior of the shadcn-ui site, to make things line up a little better.

Here is the result after (md):
Screenshot 2024-08-17 at 1 46 41 PM

And (lg):
Screenshot 2024-08-17 at 1 47 00 PM


huntabyte (approved):
Thank you!

Issue #1189, opened 2024-07-16
Enhancement: TanStack Table


jasonwillschiu:
### Describe the feature

See here
https://github.com/bryanmylee/svelte-headless-table/issues/224

Plus shadcn/ui says it's based on Tanstack Table.

Thanks for all your hard work!
Jason

huntabyte:
The last time I tried tanstack table it was a nightmare. I'll revisit the Svelte 5 rewrite and see how it goes.

jessielaf:
@huntabyte  I have been using tanstack table with Svelte for a while without any problems. Which problems did you encounter?

robots4life:
[https://github.com/TanStack/table/pull/5403](https://github.com/TanStack/table/pull/5403)

jasonwillschiu:
Happy for it to be something that's not TanStack table. Just something that's maintained with similar functionality to ease my mind haha

ESchouten:
Personally I would prefer Tanstack Table, have been using it for over a year and nothing else really comes close to it.

shyakadavis:
I'm indifferent to what ends up being used, but here is something else for consideration. (The runes-powered v2 is currently in preview)

- Repo: https://github.com/vincjo/datatables | Runes - https://github.com/vincjo/datatables/tree/runes
- URL: https://vincjo.fr/datatables.runes/

j13n:
Just tried @tanstack/svelte-table@alpha on one of my current tables since I want to start using Svelte 5. Seems to be working fine.

You do have to have `svelte@next` installed though so for now it looks like:
Svelte 4 with svelte-headless-table and Svelte 5 with @tanstack/svelte-table@alpha.

If @huntabyte and the rest are cool with it I can write some documentation on how to use it. Perhaps a new section on the bottom of the current data-table docs with instructions on how to make data-tables with Svelte 5.

robots4life:
This is just a heads up for anyone following the current docs on Data Table.

I had an [issue](https://github.com/huntabyte/shadcn-svelte/issues/1209) where I wanted to combine pagination and sort with the current Data Table per docs but use a different icon for sorting the table up and down.

The solution is to include 2 unused properties, ` pageCount` and `pageSize`  for the Data Table to work correctly with `props.sort.order`.

```ts
const { hasNextPage, hasPreviousPage, pageIndex, pageCount, pageSize } = pluginStates.page;
```

[...]

PR #1260, opened 2024-09-02
feat: Switch to TanStack Table

shyakadavis:
Closes: #1189 

I wonder if we should wait for the Svelte 5 version, though? If that's the case, happy to close this and wait for it. 🙂 

TODO:

- [ ] If/when greet lit, I can try my hand at overhauling the docs for the datable, because this P.R only switches the `Tasks` example
- [x] Center text when no search result(s)
- [x] Fix faceted filters
- [x] Clean-up change functions (`onRowSelectionChange`, `onSortingChange`, `onColumnFiltersChange`, `onColumnVisibilityChange`), preferably to separate function definitions
- [x] When you select all, and then deselect at least one row, the header's checkbox remains checked. This is because `checked` isn't reacting to changes to `table.getIsAllPageRowsSelected() || (table.getIsSomePageRowsSelected() && 'indeterminate')`. Will keep investigating.
- [x] Figure how generics play into this - I forewent using them because I kept running into weird TS errors, and used the `Task` type instead.
- [ ] Move from the `/tasks-2` route to `/tasks`
- [ ] Changed the file names to shorter/clear ones, because I often find it hard to locate a file (I know `cmd+p` is the way to go), but you know, 🤷 I can revert to the old way, though 🙂 

---

- If your PR isn't addressing a small fix (like a typo), it references an issue where it is discussed ahead of time and assigned to you. In many cases, features are absent for a reason.
- Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- This message body should clearly illustrate what problems it solves.
- Format & lint the code with `pnpm format` and `pnpm lint`


kaechele:
Nice! I started working on this in parallel but you beat me to it. My solution is very close to what you have.

Generics should technically be straightforward. Here's what works for me.
You'd add `generics="TData, TValue"` to your `<script lang="ts">` and:

- your `data` is now `TData[]` instead of `Task[]`
- `Table<Task>` becomes a `Table<TData>`
- your `ColumnDef<Task>` become a `ColumnDef<TData, TValue>`
- a `Column<Task>` is now a `Column<TData, TValue>`

One could think about pulling the column definition out of the components. This is what I ended up doing.
Composing the `Task` table specific cells into the table happens in the column definitions anyway. That way the base components become more reusable.

shyakadavis:
Hey, @kaechele 

It's my first time working with TanStack-anything, so glad I wasn't far off with how others are doing it. 😅 

> You'd add generics="TData, TValue" to your <script lang="ts">

Yup. This is what I was doing, but when it came to passing props in the column defs, weird errors came up. Which leads me to:

> One could think about pulling the column definition out of the components.

Sure. I could look into this. Thanks.

> Composing the Task table specific cells into the table happens in the column definitions anyway. That way the base components become more reusable.

Hmm... can I trouble you with a snippet of this? Can't get my head around what you mean, yet.

> EDIT: I'm actually working on this using @tanstack/svelte-table@alpha and Svelte 5, so a few things are a bit different to account for the Svelte 5 changes. In hindsight, I should have probably added a YMMV to the generics part. Haven't tested it with the older versions.

I'm actually wondering if I should just wait for Svelte 5 as well. I guess I'm doing it now in case it's easier when migrating to S5. 🤷

c-kirkeby:
In the TanStack docs they recommend using [column helpers](https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers) for building column definitions which are quite similar to the API from svelte-headless-table. I believe the rationale is that they provide a higher type safety.

entrenadorhispano:
TanStack, as soon as I understand, Will not be supported out of the box on svelte 5. 
 https://github.com/sveltejs/svelte/issues/10359#issue-2111964532
 
 Be aware of this issue.

shyakadavis:
> In the TanStack docs they recommend using [column helpers](https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers) for building column definitions

Ah. Thanks a lot, @c-kirkeby Will change this in a few.

shyakadavis:
@kaechele 

Before I commit my interpretation of this:

> One could think about pulling the column definition out of the components.

Did you mean doing something like this:

header.svelte
```svelte
	type $$Props = HTMLAttributes<HTMLDivElement> & {
		title: string;
		canSort: boolean;
		isSorted: boolean | 'asc' | 'desc' | boolean;
		toggleSorting: (desc?: boolean, isMulti?: boolean) => void;
		toggleVisibility: (v: boolean) => void;
	};
```

and in columns.ts
```ts
columnHelper.accessor('id', {
		id: 'task',
		header: ({ column }) => {
			return renderComponent(ColumnHeader, {
				title: 'Task',
				canSort: column.getCanSort(),
				isSorted: column.getIsSorted(),
				toggleSorting: column.toggleSorting,
				toggleVisibility: column.toggleVisibility
			});
		},
		enableSorting: false,
		enableHiding: false
	}),
```

c-kirkeby:
Okay so rather than doubling up on trying to convert the Tasks demo, I implemented the Data Table for my own project here at https://github.com/c-kirkeby/osrs-price-site/pull/3. It doesn't have the filter or the actions / row selection or the filters from the Tasks example, and mine has some more complex use cases like colouring cells / right-aligning cells.

One issue I ran into was this:
https://github.com/c-kirkeby/osrs-price-site/pull/3/files#diff-dfcbcef4795391b3fac61ea52dc73d867264ba430a9ea6dfcab677acd7777e66R72-R73

Due to TanStack/table#4382 there are some TypeScript errors when assigning `columns` to the `ColumnDef<TData, TValue>` type.

shyakadavis:
Closing as this has been shipped in the `next` branch. 🎉 (PR: #1318)

PR #1291, opened 2024-09-30
fix: globals.css

onlyfortesting:
Invalid CSS. But, not sure what you're trying to do


huntabyte:
What is invalid about it?

PR #1294, opened 2024-10-03
fix: Next Button Component with href

reesebre82:
When passing an href to the Button Component, the button does not redirect to the href on click. It appears that {...restProps} doesn't contain the href passed. 

Manually passed the href variable onto the a tag and links work as expected.


huntabyte (approved):
Nice catch! Thank you! Just need to run `pnpm format` at the root of the project and this will be good to merge.

PR #1310, opened 2024-10-09
patch: correct calendar direction buttons

shyakadavis:
Coincidentally, the last commit to these files was "my hand hurts". Sorry, HunterxHunter. 😅


huntabyte (approved):
Thank you!

PR #1320, opened 2024-10-13
fix: make `isActive` prop optional

kaiserinn:
Currently the `isActive` prop is required:
```ts
type $$Props = PaginationPrimitive.PageProps &
  Props & {
    isActive: boolean;
  };
```
where it should've been optional. This PR addresses that issue by changing the `isActive` to be optional.


AdrianGonz97:
@huntabyte While this change targets the `main` branch, the same issue is true in the `next` branch. I feel like `isActive` _should_ be a required prop and we should instead remove its fallback value. Thoughts?

AdrianGonz97:
Closing in favor of #2085

PR #1400, opened 2024-10-29
docs: Remove Select.Value in Select component

### LINKED ISSUE #1399: The Select component shows outdated docs.
### Change Type

Correction

### Proposed Changes

The [docs for select](https://next.shadcn-svelte.com/docs/components/select) show invalid usage of the component.

I added the select component with the CLI but there is no exported type for Select.Value like suggested in the usage. I checked the repo. The previous version does include it but not the next version

New version without the Value type: https://github.com/huntabyte/shadcn-svelte/blob/next/sites/docs/src/lib/registry/new-york/ui/select/index.ts

Old version that includes it https://github.com/huntabyte/shadcn-svelte/blob/main/sites/docs/src/lib/registry/new-york/ui/select/index.ts

shyakadavis:
Closes #1399

### DIFF
--- a/sites/docs/src/content/components/select.md
+++ b/sites/docs/src/content/components/select.md
@@ -45,9 +45,7 @@ Install `bits-ui`:
 </script>
 
 <Select.Root>
-  <Select.Trigger class="w-[180px]">
-    <Select.Value placeholder="Theme" />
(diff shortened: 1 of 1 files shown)


huntabyte (approved):
Thank you!

PR #1402, opened 2024-10-30
fix: update tabindex attributes to use numbers

shyakadavis:
Not sure why this wasn't errant before, but in a new scaffold, I was getting this error:
```
Type 'string' is not assignable to type 'number'.
```
Happy to close if this leads to unexpected side effects.

P.S. I was using `"svelte-radix": "^2.0.1"` not `"svelte-radix": "2.0.0-next.5"` maybe that's where the mismatch was coming from 🤷


huntabyte (approved):
Thank you!

PR #1405, opened 2024-10-30
docs: highlight correct line in Sonner example

shyakadavis:
I'm doing the 7GUIs tasks, ergo these patches for what I ran into🙂


huntabyte (approved):
Thank you!

PR #1414, opened 2024-10-31
fix: wrong icon import for `ScrollDownButton` in `Select`

nchaikh:
The `chevron-up` icon was being imported instead of the `chevron-down` icon


huntabyte (approved):
Thank you!

Issue #1424, opened 2024-11-01
Select in Card chapter broken (next documentation)


JanTrichter:
### Change Type

Correction

### Proposed Changes

Hi!
there's a bug in the [Card Component Documentation](https://next.shadcn-svelte.com/docs/components/card) where selecting a value for `Framework` will display `[Object object]` instead of the proper label.

How to replicate:
1. Go to the [next documentation for the card](https://next.shadcn-svelte.com/docs/components/card)
2. Set style as `Default` (this will not happen when style is `New York`)
3. Select any of the `Frameworks` and you will see the Select shows `[Object object]` instead of the proper label.

![image](https://github.com/user-attachments/assets/2e936185-0c88-4ed5-bf8f-55079eca2c1d)

Appears in both Chrome and Firefox in the Svelte 5 preview (next) documentation.
The "old" documentation works fine in both cases.

PR #1425, opened 2024-11-01
docs: fix card select showing [object object]

JanTrichter:
solves issue #1424 

- fixes selected showing `[object object]` instead of the actual name of the selected framework when using the `Default` style

Before:
image

After:
image


huntabyte (approved):
Thank you!

PR #1448, opened 2024-11-05
docs: daisyui integration

andre-brandao:
hey, i love shadcn, it was the first library I used when learning svelte, I kinda moved away from it because I love just having the utility classes daisyui provides without importing anything, but I still kinda missed some of the components from shadcn, so i took some time and integrated both of than, just wrote some documentation to help the next person that is trying to do the same

also have an exemple of how it looks with the music exemple and a theme changer on the top right

https://shad-daisy.vercel.app/
https://github.com/andre-brandao/shad-daisy

### DIFF
--- a/sites/docs/src/content/theming.md
+++ b/sites/docs/src/content/theming.md
@@ -152,24 +152,24 @@ It's recommended to use [HSL colors](https://www.smashingmagazine.com/2021/07/hs
 
 See the [Tailwind CSS documentation](https://tailwindcss.com/docs/customizing-colors#using-css-variables) for more information on using `rgb`, `rgba` or `hsl` colors.
 
-
-
 ## Integrating with other theming libraries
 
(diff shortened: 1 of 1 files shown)


andre-brandao:
Do I need to add a changeset for docs?

huntabyte:
Hey there, @andre-brandao, great job working this out!

This would be a nice thing to add to GitHub Discussions (please add this guide there) as a reference to point people to, but I'm not going to merge this into the project.

This puts the burden on us (the maintainers) to keep up with any changes to DaisyUI, which we didn't sign up for and do not wish to do. 

If you'd like to maintain this guide, you can do so via a GitHub Discussions post and we'll gladly point those interested in integrating with DaisyUI there!

Hope you understand!

andre-brandao:
@huntabyte just did that here

PR #1466, opened 2024-11-08
fix: copy button

kmalloy24:
Sorry for the rogue PR, I know this is a little more than a minor documentation/typo fix, but I figured it was easier to see the changes and pull it down to test this way. 

I only spent 15 minutes on it so I won't be offended if it isn't merged/accepted. But I was using the Hex -> Color Channel converter in /theming and noticed the copy button wasn't working.


huntabyte:
No longer needed as we don't have this component anymore, thank you!

PR #1472, opened 2024-11-09
next(breaking): replace radix icons with lucide

huntabyte:
Reference https://x.com/shadcn/status/1853902179041702169


huntabyte:
If someone wants to give this a review before I merge that would be great 😃

PR #1478, opened 2024-11-12
fix: Pin svelte-radix version for Svelte 4 support

ieedan:
Looks like [1.1.1](https://github.com/shinokada/svelte-radix/releases/tag/v1.1.1) is the latest version of radix for svelte 4.

P.S. I think the registry needs to be built but the build is broken on windows (I only have access to a windows machine rn) so you may need to build and commit.


huntabyte:
The registry rebuilds during CI before deployment which is why removed it from `next` so no worries there.

PR #1479, opened 2024-11-12
fix(next): add newline to end of components.json

kaechele:
All files generated from `templates.ts` end with a newline. `components.json` should, too.


AdrianGonz97:
thank you!

AdrianGonz97 (approved):

PR #1482, opened 2024-11-13
feat (next): Add ability disable portal for portalled components

saturnonearth:
I added the ability to set portal on dialog content (for both `Dialog` and `AlertDialog`). It used to be on `Dialog.Root` But I didn't know how to add it back there so I just added to `Dialog.Content` and `AlertDialog.Content`.

Edit: I added same props to `Popover` as well.


huntabyte:
In this case, we'd need to add it to all the floating elements, and a more flexible approach would probably be a `portalProps` type which includes all the possible props to pass to `Portal`

saturnonearth:
> In this case, we'd need to add it to all the floating elements, and a more flexible approach would probably be a `portalProps` type which includes all the possible props to pass to `Portal`

Hopefully these modifications are what you wanted?

huntabyte:
@saturnonearth yes, but let's make the name `portalProps` instead of just `portal`. Having a `portal` prop instantly makes me think it's going to be a boolean whereas `portalProps` is a bit more explicit.

huntabyte:
`DropdownMenu` and `ContextMenu` as well right?

huntabyte:
Thanks, man!  I _think_ the only one we're missing now is the `Menubar`?

huntabyte:
I'll get that type exported!

huntabyte (approved):
Thanks a ton dude! @saturnonearth

PR #1487, opened 2024-11-15
fix(next): resolve command for sveltekit sync as `execute-local`

ieedan:
Fixes #1486. 

Sorry about the misunderstanding I skipped like 2 words reading the issue and totally threw me off.

PR #1491, opened 2024-11-15
docs(next): Add `run dev server once` to SvelteKit installation instruction

### LINKED ISSUE #1474: $lib/components" does not use an existing path alias defined in your jsconfig.json
### Describe the bug

I'm using a Svelte (not SvelteKit) project managed with Vite, no typescript. When I run `npx shadcn-svelte@latest init` I get a message `$lib/components" does not use an existing path alias defined in your jsconfig.json`. 

jsconfig looks has this:

```
"paths": {
    "$components": ["src/components"],
    "$components/*": ["src/components/*"],  
    "$utils": ["src/utils"],
    "$utils/*": ["src/utils/*"] 
  }
```

vite.config has:

```
resolve: {
    alias: {
      $lib: path.resolve("./src/lib"),
      $components: path.resolve("./src/components")
    },
  }
```

Thanks!

The vite installation of Svelte only comes with $lib, I added $ components and $utils myself, perhaps init should update the config files for you?

### Reproduction

https://github.com/ianengelbrecht/svelte-shadcn-time-picker

Clone, install, and try npx shadcn-svelte@latest init

### Logs

_No response_

### System Info

```bash
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1345U
    Memory: 17.81 GB / 31.66 GB
  Binaries:
    Node: 20.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (129.0.2792.89)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    svelte: ^5.1.3 => 5.1.13
```

### Severity

blocking all usage of shadcn-svelte

ieedan:
Fixes #1474 

- Adds `run dev server once` to installation instruction 
- Adds `<PMRun/>` component
- Refactors the pm-block component with `package-manager-detector` so adding new commands is easier.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
         version: 9.12.0
       '@huntabyte/eslint-config':
         specifier: ^0.3.2
-        version: 0.3.2(@vue/compiler-sfc@3.5.12)(eslint-plugin-svelte@2.44.1(eslint@9.7.0)(svelte@5.0.4))(eslint@9.7.0)(svelte-eslint-parser@0.42.0(svelte@5.0.4))(svelte@5.0.4)(typescript@5.6.3)(vitest@2.1.3)
+        version: 0.3.2(@vue/compiler-sfc@3.5.12)(eslint-plugin-svelte@2.44.1(eslint@9.7.0)(svelte@5.0.4))(eslint@9.7.0)(svelte-eslint-parser@0.42.0(svelte@5.0.4))(svelte@5.0.4)(typescript@5.6.3)(vitest@2.1.3(@types/node@20.16.14)(terser@5.36.0))
       '@typescript-eslint/eslint-plugin':
         specifier: ^8.10.0
         version: 8.10.0(@typescript-eslint/parser@8.10.0(eslint@9.7.0)(typescript@5.6.3))(eslint@9.7.0)(typescript@5.6.3)
@@ -216,6 +216,9 @@ importers:
       mode-watcher:
         specifier: ^0.3.1
         version: 0.3.1(svelte@5.0.4)
+      package-manager-detector:
+        specifier: ^0.2.2
+        version: 0.2.2
       paneforge:
         specifier: 1.0.0-next.1
         version: 1.0.0-next.1(svelte@5.0.4)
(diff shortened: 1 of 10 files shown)


ieedan:
I almost want to open up an issue on `package-manager-detector` to see if they would be open to publishing another package with just the `resolveCommand` functionality as it is really useful for the application we are using it for now and we surely aren't and won't be the only ones to experience this.

AdrianGonz97:
I don't think this will be necessary now that #1487 is merged. `svelte-kit sync` ensures that the `.svelte-kit` directory is generated. If it's missing in the project, the CLI runs the command to generate it for them

ieedan:
> I don't think this will be necessary now that #1487 is merged. `svelte-kit sync` ensures that the `.svelte-kit` directory is generated. If it's missing in the project, the CLI runs the command to generate it for them

I think you are right let me try and reproduce their issue with the new steps before closing it!

ieedan:
Yeah I can't reproduce it with the latest version.

I still think this has some good changes to the PM components but unless something changes with `package-manager-detector` we can't really use them.

ieedan:
Changing this just to refactor the PM component. 

All I had to do to make it work was use type only imports from the root of the project so that it doesn't load `node`. And then import `resolveCommand` from `/commands`.

ieedan:
> 
> 
> https://github.com/user-attachments/assets/fc325629-0d35-42db-9ced-46097d077338
> 
> 
> 
> Is this actually how scripts are executed with bun or should it be `bunx`?

Yeah you're right I'm either using resolve command wrong or there's a bug I can hopefully find the time to take a look today if not tomorrow

ieedan:
Okay so actually bunx is a shorthand for bun x

https://bun.sh/docs/cli/bunx

But I think pretty much everyone just says bunx so maybe I should just make an exception for bun and that command?

huntabyte:
@AdrianGonz97 what are your feelings about this lol

huntabyte:
We're going to keep it as `bun x` as it is the _actual_ command rather than an alias.

huntabyte (approved):
Thank you!

PR #1494, opened 2024-11-16
docs: Fixed accordion variable names in migration guide

TheOnlyTails:
This broke the transition, since the bits-ui variable uses a different prefix.


huntabyte (approved):
Thank you!

PR #1511, opened 2024-11-21
fix: `aria-hidden` to `aria-hidden="true"` in `pagination-ellipsis`

### LINKED ISSUE #1510: pagination-ellipsis.svelte: The value of 'aria-hidden' must be either 'true' or 'false'. It cannot be empty
### Describe the bug

vite/svelte-warn prints this:

```
pagination-ellipsis.svelte:2:4 The value of 'aria-hidden' must be either 'true' or 'false'. It cannot be empty
```

### Reproduction

https://github.com/huntabyte/shadcn-svelte/blob/main/sites/docs/src/lib/registry/default/ui/pagination/pagination-ellipsis.svelte#L13

ieedan:
Fixes #1510

Did a quick search this is the only place I saw this issue in the project.

### DIFF
--- a/sites/docs/src/lib/registry/default/ui/pagination/pagination-ellipsis.svelte
+++ b/sites/docs/src/lib/registry/default/ui/pagination/pagination-ellipsis.svelte
@@ -10,7 +10,7 @@
 </script>
 
 <span
-	aria-hidden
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
Thank you!

PR #1514, opened 2024-11-22
fix(cli): Use `pathe` instead of `node:path`

ieedan:
Fixes testing issues for the CLI for windows users.

- Replaces `node:path` with `pathe`

Not sure if this is really important especially since it's an extra dependency for the CLI. But I tried it in one of my side projects and it worked without any issues.

The registry build has the exact same issue so we may be able to do the same there.

P.S.
Didn't add a changeset since there are no user facing changes lmk if I should add one anyways.


AdrianGonz97:
Thank you, but I would rather avoid adding another dependency if it's possible! The issue was that the aliases for `hooks` and `ui` weren't being normalized, so they were causing the tests to fail on windows. Opened #1516 as an alternate PR

huntabyte:
`package.json` police 👮‍♂️ strikes again! 😃 

Gonna close this in favor of #1516 - thanks for bringing it up @ieedan

PR #1543, opened 2024-12-10
docs: Remove outdated card docs

### LINKED ISSUE #1531: Card title tag not changing
### Describe the bug

I can't change the tag of the Card.Title. I'm getting a TypeScript error:

`Object literal may only specify known properties, and '"tag"' does not exist in type 'HTMLAttributes<HTMLDivElement> & { ref?: HTMLElement | null | undefined; } & { level?: 1 | 2 | 3 | 4 | 5 | 6 | undefined; }'.ts(2353)`

In Svelte 4 this worked flawlessly, but in Svelte 5 (and shadcn-svelte@next) the tag prop is not even being suggested by the IDE. Only relevant prop I have available is 'level'

Also in the docs I see this: `By default, the <Card.Title> component renders an <h3> element. `
This is not true for me, the title is using this element: `<div role="heading" aria-level="3">`

### Reproduction

Simply install the card component with `npx shadcn-svelte@next add card`

### Logs

_No response_

### System Info

```bash
Not relevant, I think
```

### Severity

annoyance

ieedan:
Fixes #1531 

Removes outdated docs for card.

### DIFF
--- a/sites/docs/src/content/components/card.md
+++ b/sites/docs/src/content/components/card.md
@@ -50,22 +50,6 @@ links:
 </Card.Root>
 ```
 
-### Modify the heading level
-
-By default, the `<Card.Title>` component renders an `<h3>` element. You can change this by passing a `tag` prop to the component.
-
-For example:
-
-```svelte
-<Card.Title tag="h1">This will render an H1</Card.Title>
-```
-
-...
-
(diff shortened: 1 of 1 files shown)


huntabyte (approved):
Thank you!

PR #1550, opened 2024-12-12
fix: Update text sizes in inputs for improved Safari support

### LINKED ISSUE #1548: Fix automatic zoom in on Safari/iOS when focusing textarea or input
### Describe the bug

Safari on iOS zooms in when the user focuses an <input> or <textarea> with font-size smaller than 16px. After the focus is lost Safari does not zoom out again. This can leave the website in a weird looking state.

When looking for solutions I only found pretty ugly hacks and it made me wonder if this issue happens with shadcn too or how they solved it. Turns out they changed from "text-sm" to "text-base md:text-sm" for Input and Textarea in [this commit](https://github.com/shadcn-ui/ui/commit/13c97acf9f6db4b3e89ec8943b6d4463dc575f85#diff-bccc7161022356d1158e1a6412faf48bbe0c296d0b2aa6146418e9dc582c8901R7)

So I propose to implement this change to solve the issue and also align the behavior of shadcn-svelte with shadcn in this regard.
I can create a PR for this, if that's ok.

### Reproduction

Not really needed I guess as this happens with every Input and Textarea on iOS & Safari

### Logs

_No response_

### System Info

```bash
System:
    OS: Windows 10 10.0.19045
    Memory: 13.40 GB / 31.34 GB
  Binaries:
    Node: 22.12.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.14.4 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Chromium (131.0.2903.51)
    Internet Explorer: 11.0.19041.4355
  npmPackages:
    svelte: ^5.10.0 => 5.10.0
```

### Severity

annoyance

ieedan:
Fixes #1548 

- Updates `<Textarea/>`, `<Input/>`, and `<Command.Input/>` to use the same styles used in the [shadcn/ui commit](https://github.com/shadcn-ui/ui/commit/13c97acf9f6db4b3e89ec8943b6d4463dc575f85#diff-bccc7161022356d1158e1a6412faf48bbe0c296d0b2aa6146418e9dc582c8901R7)
- Bumps bits-ui -> `next.71`

P.S. `<Command.Input/>` isn't in that original commit but I think it should be added anyways since it will cause the same issue

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
         version: 9.12.0
       '@huntabyte/eslint-config':
         specifier: ^0.3.2
-        version: 0.3.2(@vue/compiler-sfc@3.5.12)(eslint-plugin-svelte@2.44.1(eslint@9.7.0)(svelte@5.0.4))(eslint@9.7.0)(svelte-eslint-parser@0.42.0(svelte@5.0.4))(svelte@5.0.4)(typescript@5.6.3)(vitest@2.1.3)
+        version: 0.3.2(@vue/compiler-sfc@3.5.12)(eslint-plugin-svelte@2.44.1(eslint@9.7.0)(svelte@5.0.4))(eslint@9.7.0)(svelte-eslint-parser@0.42.0(svelte@5.0.4))(svelte@5.0.4)(typescript@5.6.3)(vitest@2.1.3(@types/node@20.16.14)(terser@5.36.0))
       '@typescript-eslint/eslint-plugin':
         specifier: ^8.10.0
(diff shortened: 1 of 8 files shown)


ieedan:
> Thanks for the fast fix! I saw you used sm:text-sm instead of md:text-sm in all but one file. I guess that is a typo?

You're right thank you for catching this!

ieedan:
All good now!

huntabyte (approved):
Thanks man!

PR #1551, opened 2024-12-12
fix(next): parse `<script module>` tags for imports

### LINKED ISSUE #1534: next: Render issue when bind:open on Select.Root
### Describe the bug

The Select component is not rendering correctly when using `bind:open`.

### Reproduction

```svelte
<Select.Root
	bind:open={selectRegionOpen}
	type="single"
	value={cluster.regionId}
	onValueChange={(value) => handleRegionChange(cluster.id, value)}
>
```

image

---

Without `bind:` it renders as expected.

image

```svelte
<Select.Root
	open={selectRegionOpen}
	type="single"
	value={cluster.regionId}
	onValueChange={(value) => handleRegionChange(cluster.id, value)}
>
```

### Logs

_No response_

### System Info

```bash
Versions:

- https://github.com/huntabyte/shadcn-svelte/tree/8b0c6d29935f6192b87f6ed5c09cbc12d436873b
- "bits-ui": "1.0.0-next.67"
```

### Severity

annoyance

ieedan:
Fixes #1534 

The root of #1534 was actually that the `index.json` and `button.json` files did not include `bits-ui` as a dependency. With a little (a lot more than I am proud of) digging I figured out that we aren't parsing the `<script module>` tags in `.svelte` files. (I did this same thing in my own project need to go fix that now).

This PR adds second walk of the `<script module>` tag if it exists so that we catch all the imports.

### DIFF
--- a/sites/docs/scripts/registry.ts
+++ b/sites/docs/scripts/registry.ts
@@ -1,6 +1,6 @@
 import * as acorn from "acorn";
 import tsPlugin from "acorn-typescript";
-import { walk } from "estree-walker";
+import { walk, type Node } from "estree-walker";
 import fs from "node:fs";
 import path from "node:path";
 import { parse, preprocess } from "svelte/compiler";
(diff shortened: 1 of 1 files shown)


ieedan:
You can prove this works by looking at the preview:
https://85a03792.shadcn-svelte.pages.dev/registry/index.json
versus the current:
https://next.shadcn-svelte.com/registry/index.json

huntabyte (approved):
Thanks man!

PR #1553, opened 2024-12-13
registry: pin deps

huntabyte:
As we release new majors for dependencies that only support Svelte 5, we'll pin those dependencies to their latest Svelte 4-supported version.

### DIFF
--- a/sites/docs/scripts/registry.ts
+++ b/sites/docs/scripts/registry.ts
@@ -156,8 +156,8 @@ async function getDependencies(filename: string, source: string) {
 	const registryDependencies = new Set<string>();
 	const dependencies = new Set<string>();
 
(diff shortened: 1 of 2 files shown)


ieedan:
Just a thought but a better way to do this may be to just use the version found in the nearest package.json. This way you don't have to maintain a list.

huntabyte:
> Just a thought but a better way to do this may be to just use the version found in the nearest package.json. This way you don't have to maintain a list.

We considered that before but then it would make more sense to do it for all the deps, which means anytime a patch release for anything comes out, we also need to update shadcn-svelte or users will have issues.

There will only be a few of these, and the previous versions of the packages will not be updated once the new major is out, so I don't mind the list.

PR #1566, opened 2024-12-19
docs(next): Fix combobox example text alignment

ieedan:
The current default example:
image

The new york example works fine since it doesn't have the `ml-auto`.

This PR just removes the `ml-auto` class.


huntabyte (approved):
Thank you!

PR #1569, opened 2024-12-20
chore(next): runed `MediaQuery` -> svelte `MediaQuery`

ieedan:
- Bump svelte to `5.7.0+` for the new `svelte/reactivity` api.
- Changed any runed `MediaQuery` references to the `svelte/reactivity` version.

Technically with this change we no longer need `runed` as a dependency of the project so not sure if you'd like to remove it or keep it for the time being!


huntabyte:
Yes, we can remove it until we need it again 😃

Thanks man!

ieedan:
Should be good to go now!

huntabyte (approved):
Thanks man!

PR #1585, opened 2024-12-28
fix(next): misplaced `className` in `cn` functions

### LINKED ISSUE #1582: Misplaced `className` in `buttonVariants`
### Describe the bug

The `className` is currently placed in `buttonVariants` instead of `cn` function.

Current:
```ts
<ButtonPrimitive.Root
	{builders}
	class={cn(buttonVariants({ variant, size, className }))}
                                                  ^
	// more codes here…
>
```

Should be:
```ts
<ButtonPrimitive.Root
	{builders}
	class={cn(buttonVariants({ variant, size }), className)}
                                                     ^

	// more codes here…
>
```

### Reproduction

The easiest way to find the issue is by using Typescript error on the [Button.svelte file](https://github.com/huntabyte/shadcn-svelte/blob/main/sites/docs/src/lib/registry/new-york/ui/button/button.svelte).

### Logs

_No response_

### System Info

```bash
Not available
```

### Severity

annoyance

ieedan:
Fixes #1582

This has always been an issue on next but I think it was only now noticed after the types for the `class` prop changed in [5.16.0](https://github.com/sveltejs/svelte/releases/tag/svelte%405.16.0).

## Also
- Bumps deps
- Runs `app-state` migration since `$page` will now show deprecated warning

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
 		"prettier-plugin-svelte": "^3.2.7",
 		"prettier-plugin-tailwindcss": "^0.6.8",
 		"pretty-quick": "^4.0.0",
-		"svelte": "^5.15.0",
+		"svelte": "^5.16.0",
 		"svelte-eslint-parser": "^0.42.0",
 		"typescript": "^5.6.3",
 		"typescript-eslint": "^8.10.0"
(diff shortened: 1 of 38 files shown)


huntabyte (approved):
You're incredible.

Thank you!

Issue #1587, opened 2024-12-29
Range Calendar - background of cell containing highlighted selection-start incorrect


PickleyD:
### Describe the bug

A minor visual bug on the Range Calendar component. When an initial selection occurs on the Range Calendar, the background colour of the parent cell is changed before the end selection has occurred. The background colour should only change after the end selection has occurred. 

### Reproduction

The background colour is subtle in the default styling but can be observed in the docs:

<img width="249" alt="Screenshot 2024-12-29 at 15 08 26" src="https://github.com/user-attachments/assets/b57254ac-e170-49d1-8e8e-648bf6f6c9e0" />

The issue became more noticeable after customising the styling for my own use case.

To fix this locally I made the following change in `range-calendar-cell`:

```diff
- [&:has([data-selected])]:bg-accent
+ [&:has([data-selected]):not(:has([data-highlighted]))]:bg-accent
```

### System Info

```bash
N/A
```

### Severity

annoyance

PickleyD:
Writing to volunteer to submit a PR for this with the above mentioned fix

PR #1605, opened 2025-01-08
fix: disable bg-accent from range-calendar-cell when it is the highlighted element

PickleyD:
Resolves https://github.com/huntabyte/shadcn-svelte/issues/1587


huntabyte (approved):
Thank you!

PR #1606, opened 2025-01-08
docs: Update the installation documentation for svelte shadcn/ui when using vite.

rxliuli:
In the latest version of Vite, tsconfig.json has been split into three parts.

- tsconfig.app.json
- tsconfig.json
- tsconfig.node.json

Both tsconfig.app.json and tsconfig.json need to configure path aliases. In fact, the React version of shadcn/ui has already fixed this issue, so the documentation error also needs to be fixed in the Svelte version.

image
image


huntabyte (approved):
Thank you!

Issue #1615, opened 2025-01-12
next: Refactor `is-mobile` hook to use new `MediaQuery` class


Wombosvideo:
### Prerequisites

- [ ] This feature already exists in [shadcn/ui](https://github.com/shadcn/ui) - if not, it won't be considered here so don't continue with your issue.

### Describe the feature

Hey, since the Svelte team added the [`MediaQuery`](https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery) class, the `IsMobile` hook is now somewhat redundant. 

To keep compatibility with all dependent components, I suggest extending the `MediaQuery` class:

```typescript
import { MediaQuery } from "svelte/reactivity";

const MOBILE_BREAKPOINT = 768;

export class IsMobile extends MediaQuery {
	constructor() {
		super(`(max-width: ${MOBILE_BREAKPOINT - 1}px)`);
	}
}
```

Notice that this might be a breaking change for users refusing to update their dependencies.

But since `next` is still `next` and not `main`, this change is acceptable in my opinion.

Janaka-Steph:
I was wondering what was this new file when installing the new sidebar component. Not much info out there.

Wombosvideo:
This file is only used in the Sidebar components, as far as I know. `MediaQuery` creates a reactive boolean state which updates its value whenever the evaluation of the media query's condition changes. This allows you, for example, to render a button only if the viewport has a certain width, or to apply classes if the user's system prefers dark mode. The `is-mobile` hook uses this feature to determine whether the user uses a mobile device, using the viewport width.

Janaka-Steph:
This file is created in src/lib/hooks/is-mobile.svelte.ts when installing the sidebar component. This is confusing.

PR #1616, opened 2025-01-12
fix(next): update `is-mobile` hook to use new `MediaQuery` class

Wombosvideo:
## Quick Glance

- Fixes code redundancy.
- Does not introduce breaking changes for dependent components.
- Introduces breaking change when using `svelte` version older than `5.7.0`.
- Addresses #1615.

## Description

Hey, since the Svelte team added the [`MediaQuery`](https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery) class, the `IsMobile` hook is now somewhat redundant. 

To keep compatibility with all dependent components, the `IsMobile` class still exists but extends the `MediaQuery` class, which provides the same functionality.

### DIFF
--- a/sites/docs/src/lib/registry/default/hook/is-mobile.svelte.ts
+++ b/sites/docs/src/lib/registry/default/hook/is-mobile.svelte.ts
@@ -1,27 +1,9 @@
-import { untrack } from "svelte";
+import { MediaQuery } from "svelte/reactivity";
 
 const MOBILE_BREAKPOINT = 768;
 
-export class IsMobile {
-	#current = $state<boolean>(false);
-
+export class IsMobile extends MediaQuery {
 	constructor() {
-		$effect(() => {
-			return untrack(() => {
-				const mql = window.matchMedia(`(max-width: ${MOBILE_BREAKPOINT - 1}px)`);
-				const onChange = () => {
-					this.#current = window.innerWidth < MOBILE_BREAKPOINT;
-				};
-				mql.addEventListener("change", onChange);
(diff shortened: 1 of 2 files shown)


sites/docs/src/lib/registry/default/hook/is-mobile.svelte.ts
```diff
@@ -1,27 +1,9 @@
...
-				return () => {
-					mql.removeEventListener("change", onChange);
-				};
-			});
-		});
-	}
-
-	get current() {
-		return this.#current;
+		super(`(max-width: ${MOBILE_BREAKPOINT - 1}px)`);
```
huntabyte, line 7: I don't believe Svelte's `MediaQuery` expects you to wrap it in parenthesis, right?

Wombosvideo: While Svelte does wrap the query with parentheses and thus doesn't require you to write them yourself, it does not make a difference if it's wrapped twice due to how media queries are evaluated, according to [Media Queries @ MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#inverting_a_querys_meaning)

huntabyte: Got it! I still think we should just remove the parens then here to align with Svelte's example usage in their docs for `MediaQuery`!

huntabyte (approved):
Thank you! Awesome job!

PR #1617, opened 2025-01-12
docs(next): use `add` instead of `install` for `<PMInstall/>`

### LINKED ISSUE #1614: wrong yarn cmd in next.shadcn-svelte.com
### Change Type

Correction

### Proposed Changes

in the new Svelte 5 Docs the yarn cmd is `yarn install ...` instead of the correct `yarn add ...` 

if you try to run it you get a warning by yarn and it won't install 

example:
```bash
yarn install v1.22.15
error `install` has been replaced with `add` to add new dependencies. Run "yarn add @tanstack/table-core" instead.
```

the issue seems to be related with the usage of `package-manager-detector/commands` in [lib/stores/package-manager.ts](https://github.com/huntabyte/shadcn-svelte/blob/401b857a24ee2693d43bb4f04ce04268af7fba49/sites/docs/src/lib/stores/package-manager.ts)

ieedan:
Fixes #1614 

I know I will have to fix this in some of my other projects as well 🤦

### DIFF
--- a/sites/docs/src/lib/components/docs/pm-install.svelte
+++ b/sites/docs/src/lib/components/docs/pm-install.svelte
@@ -3,4 +3,4 @@
 	export let command: string = "";
 </script>
 
-<PMBlock type="install" {command} />
+<PMBlock type="add" {command} />


huntabyte (approved):
Thank you!

PR #1618, opened 2025-01-12
fix: added svg css to command-link-item

witer33:
Command.Item had the new css for svg icons, but Command.LinkItem didn't.


huntabyte (approved):
Thank you!

PR #1632, opened 2025-01-18
fix(docs): Pass `data` instead of `form.data`

### LINKED ISSUE #1630: form example <SettingsForm data={data.form} /> wrong use <SettingsForm {data} />
### Change Type

Correction

### Proposed Changes

Hi simple replace the :
```
<SettingsForm data={data.form} />
```
with 
```
<SettingsForm {data} />
```

# Why TR;TL

if you apply the docs will have this issue :

```ts
Property 'form' is missing in type 'SuperValidated<{ username: string; }, any, { username: string; }>' but required in type '{ form: SuperValidated<{ username: string; }>; }'.ts(2741)
```

in src\routes\settings\+page.svelte
```svelte
<script lang="ts">
	import SettingsForm from './settings-form.svelte';
	let { data } = $props();
	console.log(data);

{
  form: {
    id: '16btw7k',
    valid: false,
    posted: false,
    errors: {},
    data: { username: '' },
    constraints: { username: [Object] }
  }
}
	
</script>

<SettingsForm data={data.form} />
```

so if you pass the form then in the component src\routes\settings\settings-form.svelte  YOU USE - > data.form 

```svelte
 const form = superForm(data.form, {
     validators: zodClient(formSchema),
    });
```

you will face this issue:
```
SuperFormError: No form data sent to superForm. Make sure the output from superValidate is used (usually data.form) and that it's not null or undefined. Alternatively, an object with default values for the form can also be used, but then constraints won't be available.
```

ieedan:
Fixes #1630

### DIFF
--- a/sites/docs/src/content/components/form.md
+++ b/sites/docs/src/content/components/form.md
@@ -158,7 +158,7 @@ We'll pass the `form` from the data returned from the load function to the form 
   let { data }: { data: PageData } = $props();
 </script>
 
-<SettingsForm data={data.form} />
+<SettingsForm {data} />
 ```
 
 ### Create an Action that handles the form submission


ieedan:
No idea why it say's it failing when you go to the logs everything looks good...

huntabyte (approved):
Thanks man!

PR #1633, opened 2025-01-19
fix: fixed imports and removed useless css

witer33:
size-4 is already applied by Button and renderSnippet wasn't used in the code.


huntabyte (approved):
Thank you!

Issue #1642, opened 2025-01-24
Cli: upgrade package-manager-detector to 0.2.8 to improve bun support


BrandonTing:
### Describe the bug

Hi,
Thank you for all the hard works. This is an annoying issue but not affecting functionality.
When I try to install new components using bun 1.2, the cli tool keeps asking "Which package manager do you want to use?" rather than auto-choose bun. After dig into the source code, it looks like bun.lock is not supported before package-manager-detector v.0.2.8.
Willing to help if needed! 
![Image](https://github.com/user-attachments/assets/58213864-304c-4202-b720-26b47ea1cd2e)

### Reproduction

https://github.com/BrandonTing/svelte-shadcn-repro

### Logs

```bash

```

### System Info

```bash
System:
    OS: macOS 13.4
    CPU: (8) arm64 Apple M2
    Memory: 140.16 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.9.2 - /usr/local/bin/npm
    pnpm: 9.0.2 - /usr/local/bin/pnpm
    bun: 1.2.0 - /usr/local/bin/bun
  Browsers:
    Chrome: 131.0.6778.265
    Safari: 16.5
  npmPackages:
    @sveltejs/kit: ^2.16.0 => 2.16.1 
    bits-ui: ^0.22.0 => 0.22.0 
    lucide-svelte: ^0.474.0 => 0.474.0 
    mode-watcher: ^0.5.0 => 0.5.0 
    svelte: ^5.0.0 => 5.19.2 
    svelte-sonner: ^0.3.28 => 0.3.28
```

### Severity

annoyance

ieedan:
Go for it!

artemszelenov:
@ieedan Hi! how is it going? there is still an issue

BrandonTing:
@artemszelenov 
sorry, in prev MR I saw @huntabyte  step in and I thought he's gonna take control since then. 
Now format issue should be solved and wait for approval

PR #1646, opened 2025-01-25
Chore bump package manager detector

BrandonTing:
Closes: #1642

bun 1.2 introduces new lock format, which is not supported by  package-manager-detector until 0.2.8. 

- If your PR isn't addressing a small fix (like a typo), it references an issue where it is discussed ahead of time and assigned to you. In many cases, features are absent for a reason.
- Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- This message body should clearly illustrate what problems it solves.
- Format & lint the code with `pnpm format` and `pnpm lint`

### DIFF
--- a/packages/cli/src/commands/add.ts
+++ b/packages/cli/src/commands/add.ts
@@ -5,6 +5,7 @@ import color from "chalk";
 import { Command } from "commander";
 import { execa } from "execa";
 import * as v from "valibot";
+import { resolveCommand } from "package-manager-detector";
 import { type Config, getConfig } from "../utils/get-config.js";
 import { getEnvProxy } from "../utils/get-env-proxy.js";
 import { ConfigError, error, handleError } from "../utils/errors.js";
@@ -244,19 +245,25 @@ async function runAdd(cwd: string, config: Config, options: AddOptions) {
 	}
 
 	// Install dependencies.
-	const commands = await detectPM(cwd, options.deps);
-	if (commands) {
-		const [pm, add] = commands.add.split(" ") as [string, string];
-		tasks.push({
-			title: `${highlight(pm)}: Installing dependencies`,
-			enabled: dependencies.size > 0,
-			async task() {
-				await execa(pm, [add, "-D", ...dependencies], {
-					cwd,
-				});
-				return `Dependencies installed with ${highlight(pm)}`;
-			},
-		});
+	const pm = await detectPM(cwd, options.deps);
+	if (pm) {
+    const add = resolveCommand(pm, "add", ["-D", ...dependencies]);
+    if(add) {
+      tasks.push({
+        title: `${highlight(pm)}: Installing dependencies`,
+        enabled: dependencies.size > 0,
+        async task() {
+          await execa(add.command, add.args, {
+            cwd,
+          });
+          return `Dependencies installed with ${highlight(pm)}`;
+        },
+      });
+    } else {
+      p.log.warn(
+        `Could not detect a package manager in ${cwd}.`
+      );  
+    }
 	}
 
 	await p.tasks(tasks);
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/add.ts
```diff
@@ -244,19 +245,23 @@ async function runAdd(cwd: string, config: Config, options: AddOptions) {
...
+				enabled: dependencies.size > 0,
+				async task() {
+					await execa(add.command, add.args, {
+						cwd,
+					});
+					return `Dependencies installed with ${highlight(pm)}`;
+				},
+			});
+		} else {
+			p.log.warn(`Could not detect a package manager in ${cwd}.`);
```
ieedan, line 263: On `@next` we just throw an error since if add wasn't resolved it was due to incorrect code.

https://github.com/huntabyte/shadcn-svelte/blob/next/packages/cli/src/commands/add.ts#L292

packages/cli/src/commands/init.ts
```diff
@@ -363,19 +364,23 @@ export async function runInit(cwd: string, config: Config, options: InitOptions)
...
+				enabled: options.deps,
+				async task() {
+					await execa(add.command, add.args, {
+						cwd,
+					});
+					return `Dependencies installed with ${highlight(pm)}`;
+				},
+			});
+		} else {
+			p.log.warn(`Could not detect a package manager in ${cwd}.`);
```
ieedan, line 382: same thing here

packages/cli/src/commands/update.ts
```diff
@@ -225,19 +226,23 @@ async function runUpdate(cwd: string, config: Config, options: UpdateOptions) {
...
+				enabled: dependencies.size > 0,
+				async task() {
+					await execa(add.command, add.args, {
+						cwd,
+					});
+					return `Dependencies installed with ${highlight(pm)}`;
+				},
+			});
+		} else {
+			p.log.warn(`Could not detect a package manager in ${cwd}.`);
```
ieedan, line 244: same thing here

packages/cli/src/utils/auto-detect.ts
```diff
@@ -2,8 +2,8 @@ import fs from "node:fs";
 import path from "node:path";
 import ignore, { type Ignore } from "ignore";
 import { type TsConfigResult, getTsconfig } from "get-tsconfig";
-import { detect } from "package-manager-detector";
-import { AGENTS, type Agent, COMMANDS } from "package-manager-detector/agents";
+import { AGENTS, type Agent, COMMANDS, detect } from "package-manager-detector";
+// import { AGENTS, type Agent, COMMANDS } from "package-manager-detector/agents";
```
ieedan, line 6: just delete this no need to comment it

ieedan, line 6:
```suggestion
```

ieedan: You can just delete this no need to keep it around

BrandonTing: thanks! auto format somehow not working as expected. updated!

packages/cli/src/commands/add.ts
BrandonTing: thanks for the instruction. I have updated those files into throwing error.

ieedan:
LGTM!

fjeremic:
@huntabyte would it be possible cherry-pick this change into the @next version? shadcn-svelte@next uses package-manager-detector v0.2.2 which doesn't have support for detecting Deno, while the version in this PR (v1.2.0) does. When trying to install components using the shadcn-svelte@next cli, it fails to detect the Deno runtime, and it won't install component dependencies because of the code here:

https://github.com/huntabyte/shadcn-svelte/blob/95ffa008cfec6ddddbd18dcf8a8c133450dff1ec/packages/cli/src/commands/init.ts#L367-L378

Backporting this PR to the @next branch will fix the issue.

PR #1664, opened 2025-02-04
docs: update install command

rxliuli:
The `pnpm add` in the current document is not a valid command and should be changed to `pnpm i` to install dependencies.

Before
image
After
image


huntabyte (approved):
Thank you!

PR #1676, opened 2025-02-09
fix: removed duplicate avatar on themes page

oliverrr123:
### Removed duplicate avatar on themes page:
image


huntabyte (approved):
Thank you!

PR #1698, opened 2025-02-17
chore: pin sv version and remove unnecessary installation step

### LINKED ISSUE #1687: Update installation docs
### Change Type

Correction

### Proposed Changes

The installation docs for sveltekit need an update
- `npm create svelte@latest` should be `npx sv create`
- `npm dlx  sv add tailwindcss` should be `npm install tailwindcss @tailwindcss/vite` (I think)
- `npm install` is not necessary anymore

happy to write a pull request

ieedan:
Fixes #1687

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Build site
         env:
-          NODE_OPTIONS: "--max-old-space-size=8192"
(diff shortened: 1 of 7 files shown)


ieedan:
Just realized create svelte should also be pinned since 7.0.0 is deprecated. Should we pin it or just leave it alone on the basis that new projects shouldn't be started with create svelte?

huntabyte (approved):
Thank you!

PR #1699, opened 2025-02-17
chore(cli)(next): Add peerDependency precondition to CLI

ieedan:
Adds a check before running any of the commands to see if the supported versions of svelte and tailwindcss are being used if not it logs a warning.

Hopefully this will help with any confusion regarding support for tailwind v4.

Edit (Updated Preview):
image

Notes:
- This does add a semver dependency but I think it's probably a net benefit. Maybe in the future this would allow me to open a PR to prevent installing dependencies when the user already has a newer version of a dependency.


ieedan:
Thanks @AdrianGonz97!

ieedan:
> This was a great idea. Thank you!!

Sorry the implementation was so sloppy lol

huntabyte:
> Sorry the implementation was so sloppy lol

He makes me feel this way everytime he reviews my PRs too don't feel bad 😂

PR #1703, opened 2025-02-18
fix(next): remove rounded-full from avatar-fallback + use size instead of w and h

ieedan:
This PR removes the rounded-full class from the avatar-fallback implementation. 

Since we are already using overflow-hidden on `<Avatar.Root/>` there isn't really any reason this should be rounded especially since it can cause issues like the following:
```svelte
<Avatar.Root class="rounded-md">
	<Avatar.Image src=""/>
	<Avatar.Fallback>
		<Upload/>
	</Avatar.Fallback>
</Avatar.Root>
```

image

Went ahead and also changed the sizing classes to use size-full instead of w and h.


huntabyte (approved):
Thank you!

PR #1710, opened 2025-02-20
fix missing var()

nathancahill:
In some of the computed classes (but not most of them) `var()` is missing around the CSS variable.


dev000000:
tailwind v3 does require var() wrapper

huntabyte (approved):
Thank you!

PR #1722, opened 2025-02-27
fix(drawer): add bottom safe inset

temich:
When an app is installed as a PWA in standalone mode, the bottommost button overlaps with the iPhone’s "home bar".

Before:
IMG_2566

After:
IMG_2567


huntabyte (approved):
Thank you!

AdrianGonz97:

PR #1730, opened 2025-03-02
chore: Add tailwindcss to system info command in issue template

ieedan:
This should help more easily diagnose issues with incompatible tailwind versions.


huntabyte (approved):
Thank you!

PR #1741, opened 2025-03-06
docs: Specify Drawer target devices to avoid confusion

x-N0:
Drawer is only for mobile and tablets, specified to avoid confusion for people coming from other UI frameworks in which Drawers can be used in other devices, including PC.


huntabyte:
That's not true. The drawer is supported on Desktop as well.

x-N0:
@huntabyte Did not work for me on Brave, tho, but I guess you double checked it; I will discard the PR.

Issue #1745, opened 2025-03-07
◐  Initializing project..  [CLI Error]: Failed to fetch base color from registry.


Kaidstor:
### Describe the bug

when starting initialize project it cant get the registry and https://next.shadcn-svelte.com/registry response with 404 

same issue on adding 
bun x shadcn-svelte@next add button

   shadcn-svelte  v1.0.0-next.6 
│
▲  Incompatible dependency versions detected:
│  
│    need tailwindcss@3.x.x  >>  found 4.0.12
│  
│  Use at your own risk!
│
└  [CLI Error]: Failed to fetch components from registry.

### Reproduction

commands:

bun x shadcn-svelte@next init
bun x shadcn-svelte@next add button

workspace (tauri 2, empty proj) with bun

components.json
```json
{
	"$schema": "https://next.shadcn-svelte.com/schema.json",
	"style": "default",
	"tailwind": {
		"config": "tailwind.config.js",
		"css": "src/app.css",
		"baseColor": "stone"
	},
	"aliases": {
		"components": "$lib/components",
		"utils": "$lib/utils",
		"ui": "$lib/components/ui",
		"hooks": "$lib/hooks"
	},
	"typescript": true,
	"registry": "https://next.shadcn-svelte.com/registry"
}
```

### Logs

```bash

```

### System Info

```bash
System:
    OS: macOS 15.3.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 279.22 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.local/state/fnm_multishells/3943_1741340700466/bin/node
    npm: 10.9.2 - ~/.local/state/fnm_multishells/3943_1741340700466/bin/npm
    pnpm: 10.2.0 - /opt/homebrew/bin/pnpm
    bun: 1.2.1 - ~/.bun/bin/bun
  Browsers:
    Safari: 18.3
  npmPackages:
    @sveltejs/kit: ^2.9.0 => 2.18.0 
    bits-ui: ^1.3.6 => 1.3.6 
    svelte: ^5.0.0 => 5.22.2 
    tailwindcss: ^4.0.12 => 4.0.12
```

### Severity

annoyance

ieedan:
So [next.shadcn-svelte.com/registry](https://next.shadcn-svelte.com/registry) is just a basepath for a registry so it will always show a 404 since it isn't intended to exist.

[...]

PR #1746, opened 2025-03-07
feat: improve error messages when fetching registry

ieedan:
Should help diagnose issues like #1745 

Now it properly bubbles the produced errors all the way up + it generates more specific error messages when fetching / parsing json responses.


huntabyte (approved):
This will be very helpful. Thank you!

PR #1756, opened 2025-03-13
docs: Add default value to fix `calendar-with-selects` example

### LINKED ISSUE #1751: The advanced customization (Month & Year Selects) Example is brocken in preview docs
### Change Type

Correction

### Proposed Changes

Hey, I noticed that the example on the pre-release page doesn't work correctly—it doesn’t behave the same way as in the regular docs.  
The days don’t change when you click on them; the cursor always stays on the current day.  
If you try to use the code from the page, it doesn’t work either because it behaves the same way as on the site.  
Unfortunately, I haven't found a fix for this yet.  

Thanks for the great work! :)

---
Old i can select the 19

Image

new i cant select any day

Image

ieedan:
Fixes #1751

### DIFF
--- a/sites/docs/src/lib/registry/default/example/calendar-with-selects.svelte
+++ b/sites/docs/src/lib/registry/default/example/calendar-with-selects.svelte
@@ -126,7 +126,7 @@
 							<Calendar.GridRow class="mt-2 w-full">
 								{#each weekDates as date}
 									<Calendar.Cell {date} month={month.value}>
-										<Calendar.Day/>
+										<Calendar.Day />
 									</Calendar.Cell>
 								{/each}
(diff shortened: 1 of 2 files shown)


huntabyte:
Hmm, this shouldn't be the case 🤔

ieedan:
> Hmm, this shouldn't be the case 🤔

Like it shouldn't require the default value?

huntabyte:
I got confused for a second and thought this was the actual calendar component you were editing due to the exposed props!

huntabyte:
Fixed in #1758 - root cause being we were missing a `type` and by passing a value you basically tricked the internals into knowing it is a `single` type

PR #1757, opened 2025-03-13
docs: make windows work

ieedan:
This could be done other ways but it works and now I can actually build the registry :)


huntabyte (approved):
Thank you!

PR #1762, opened 2025-03-15
fix(cli)(next): add missing ui alias option

Wombosvideo:
Adds missing option `--ui-alias <path>` to CLI

This would allow a non-interactive initialization:

```bash
pnpm dlx shadcn-svelte@next init \
  --style default \
  --base-color slate \
  --css src/app.css \
  --tailwind-config tailwind.config.ts \
  --components-alias '$lib/components' \
  --utils-alias '$lib/utils' \
  --hooks-alias '$lib/hooks' \
  --ui-alias '$lib/components/ui'
```

### DIFF
--- a/.changeset/flat-sloths-learn.md
+++ b/.changeset/flat-sloths-learn.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix(cli)(next): Add missing ui alias option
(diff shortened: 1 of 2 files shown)


Wombosvideo:
I don't know how changesets work with `next` tags, so you might want to add one yourself.

ieedan:
Hey thanks for giving this a shot!

You added the option to commander but you haven't implemented it so that it actually does anything at the moment.

Wombosvideo:
Well, I added this line locally, ran `build` and tried it and it seems to already work that way. Do you know what other places I need to add it?

ieedan:
Ah my bad yeah somehow it wasn't added when this was originally implemented. LGTM!

huntabyte (approved):

PR #1768, opened 2025-03-17
chore: Upgrade to `eslint-plugin-svelte@v3`

ieedan:
The `v3` plugin introduced some new rules we should probably be enforcing on the registry to prevent users getting errors in their projects.

This bumps all the workspace root dependencies and fixes any violations of the new rules (mostly each key stuff)

Right now we have the `projectService` option (don't ask me what it does) commented out. When we have it enabled the project takes ~3 minutes to lint see: https://github.com/sveltejs/eslint-plugin-svelte/issues/1084 It does enable some other rules which found a few issues which I did fix before turning it off.

Related: #1731 

**Also**:
- Removed `svelte-eslint-parser` because I don't think we need it installed
- I have `svelte/no-useless-mustaches` set to `warn` atm because there are a few spots where it's wrong


huntabyte (approved):
Awesome, thank you!

PR #1771, opened 2025-03-18
feat(v4): Update CLI for tailwindcss@v4

ieedan:
Updates the CLI and registry for the new v4 release.

- Removes the `new-york` registry. As now `default` is the only registry.
   - Updates build-registry to now build a registry to `/registry/[ts|js]`
   - Updates all blocks examples etc so that the site works without both registries
- Removes old js based tailwind config
- Initial updates to the tailwind config for tailwind v4 support


ieedan:
This is ready to go for now

PR #1778, opened 2025-03-19
feat(v4): Strip types through the CLI

ieedan:
This PR removes the JS registry in favor of stripping the types with the CLI.

- Updates `build-registry` to only build a typescript registry under `/registry/`
- Updates the CLI to strip types at the same time as it updates the imports 
- We now use [sv-strip](https://github.com/ieedan/sv-strip) to give us better test coverage and remove some complexity

This is some pre-work for remote components as this will simplify things and also give users more features for their registries.

## CLI
We have to add a few dependencies to make this work obviously
- `ts-blank-space`
- `sv-strip`
- `prettier`
- `prettier-plugin-svelte`

It's worth noting that you could try and eliminate `prettier-plugin-svelte` and instead load the users config but you will run into issues when operating out of the current working directory. (I have this problem in jsrepo) Basically prettier uses a `process.cwd()` instead of letting you provide one so it won't work when `--cwd` is passed.


ieedan:
This should be good to go

huntabyte (approved):

PR #1783, opened 2025-03-20
chore(v4): Bundle mending

ieedan:
In #1778 I introduced some extra dependencies this PR is an attempt to remove a few of them.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,7 +48,7 @@
 		"node-fetch-native": "^1.6.4",
 		"semver": "^7.7.1",
 		"sucrase": "^3.35.0",
-		"sv-strip": "^0.0.7"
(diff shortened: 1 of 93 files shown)


ieedan:
Found another edge case with the button component gonna go back to `sv-strip` and see if I can't fix it:

```svelte
<script module>
	export const buttonVariants = tv({
		// --snip--
	});

</script>

<!-- --snip-- -->
```

ieedan:
Everything works pretty good now the one issue I am having now is https://github.com/ieedan/sv-strip/issues/24

I am not quite sure how to solve this at the moment though and the output isn't extra whitespace but just sometimes reduced whitespace so I think I am okay to merge without this and update in the future.

huntabyte (approved):

PR #1791, opened 2025-03-25
chore: tailwind v4 styles

shyakadavis:
- [x] Label
- [x] Menubar
- [x] Pagination


ieedan:
@shyakadavis just taking a first pass these are the first few things I noticed

It looks like they moved the indicator for select to the right so we need to update as well.

```diff
<SelectPrimitive.Item
	bind:ref
	{value}
	data-slot="select-item"
	class={cn(
		"focus:bg-accent focus:text-accent-foreground [&_svg:not([class*='text-'])]:text-muted-foreground outline-hidden *:[span]:last:flex *:[span]:last:items-center *:[span]:last:gap-2 relative flex w-full cursor-default select-none items-center gap-2 rounded-sm py-1.5 pl-2 pr-8 text-sm data-[disabled]:pointer-events-none data-[disabled]:opacity-50 [&_svg:not([class*='size-'])]:size-4 [&_svg]:pointer-events-none [&_svg]:shrink-0",
		className
	)}
	{...restProps}
>
	{#snippet children({ selected, highlighted })}
-		<span class="absolute left-2 flex size-3.5 items-center justify-center">
+		<span class="absolute right-2 flex size-3.5 items-center justify-center">
			{#if selected}
				<Check class="size-4" />
			{/if}
		</span>
		{#if childrenProp}
			{@render childrenProp({ selected, highlighted })}
		{:else}
			{label || value}
		{/if}
	{/snippet}
</SelectPrimitive.Item>
```

`account-switcher` example should be updated:
```diff
<Select.Trigger
		class={cn(
-			"flex items-center gap-2 [&>span]:line-clamp-1 [&>span]:flex [&>span]:w-full [&>span]:items-center [&>span]:gap-1 [&>span]:truncate [&_svg]:h-4 [&_svg]:w-4 [&_svg]:shrink-0",
+			"flex items-center w-full gap-2 [&>span]:line-clamp-1 [&>span]:flex [&>span]:w-full [&>span]:items-center [&>span]:gap-1 [&>span]:truncate [&_svg]:h-4 [&_svg]:w-4 [&_svg]:shrink-0",
			isCollapsed &&
				"flex h-9 w-9 shrink-0 items-center justify-center p-0 [&>span]:w-auto [&>svg]:hidden"
		)}
		aria-label="Select account"
	>
		<span class="pointer-events-none">
			<SelectedIcon class={cn(isCollapsed ? "" : "")} />
			<span class={cn("ml-2", isCollapsed && "hidden")}>
				{selectedAccount?.label}
			</span>
		</span>
	</Select.Trigger>
-	<Select.Content>
+	<Select.Content class="min-w-[var(--bits-select-anchor-width)]">
		{#each accounts as account (account)}
			<Select.Item value={account.email} label={account.label}>
```

ieedan:
>  Error: Property 'resize' does not exist on type '{ $on?(type: string, callback: (e: any) => void): () => void; $set?(props: Partial<PaneProps>): void; }'. (ts)
 				if (resizablePaneRef) {
 					resizablePaneRef.resize(parseInt(value));
 				}
 				
I think this is because the `export const` api calls are not passed through when re-exporting the component. I think maybe we should just not have the `data-slot` here since we aren't actually going to be able to put it everywhere anyways. And just go back to exporting { Pane } from 'paneforge'

huntabyte:
For Pagination:
>  Expression produces a union type that is too complex to represent.

This is likely a conflict with the type of `children` / `child` here. I can take a look sometime soon!

shyakadavis:
> Also the shadcn examples pretty much all show the content being the same size as the trigger for select so maybe that should be left in?

Yeah. Had this before, and commented it out due to the `popper` issue mentioned above. Maybe Hunter can advise if he is going to add it in bits-ui, if not, we can forego its check and just apply the "sameWidth" classes here.

```svelte
<!-- TODO: Figure out how  position: "popper" fits in all this as it doesn't exist in bits-ui -->
	<SelectPrimitive.Content
			// position === "popper" &&
			// "data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1",
		)}
	>
		<SelectPrimitive.Viewport
			class={cn(
				"p-1"
				// position === "popper" &&
				// 	"h-[var(--bits-select-anchor-height)] w-full min-w-[var(--bits-select-anchor-width)] scroll-my-1"
			)}
		>
			{@render children?.()}
		</SelectPrimitive.Viewport>
	</SelectPrimitive.Content>
```

huntabyte:
Yeah everything we do is `position: "popper"` so those styles should be the ones we apply 😃

huntabyte:
Is this good for review @shyakadavis ?

shyakadavis:
Hey, @huntabyte . I would say "yes", but would appreciate some guidance and/or eyes on these remaining issues:

- **Resizable**: The horizontal resize bar in not visible
- These won't accept data-slots:
	- Select.Root
	- Sheet.Root
	- Sheet.Portal
	- Tooltip.Root
	- Tooltip.Provider
- Tooltip.Arrow is not rendering as intended.

ieedan:
> These won't accept data-slots:
Select.Root
Sheet.Root
Sheet.Portal
Tooltip.Root
Tooltip.Provider

I think we had agreed these weren't necessary since they technically aren't part of the DOM.

As for the other 2 I am taking a look right now...

ieedan:
CleanShot 2025-03-31 at 08 12 20

🤔

ieedan:
https://github.com/user-attachments/assets/311e3631-5186-4ad4-bc98-07b3f229c24b

This may need some tuning but I was able to do this:
```svelte
<script lang="ts">
	import { Tooltip as TooltipPrimitive } from "bits-ui";
	import { cn } from "$lib/utils.js";

	let {
		ref = $bindable(null),
		class: className,
		sideOffset = 0,
		side,
		children,
		...restProps
	}: TooltipPrimitive.ContentProps = $props();
</script>

<TooltipPrimitive.Portal>
	<TooltipPrimitive.Content
		bind:ref
		data-slot="tooltip-content"
		{sideOffset}
		{side}
		class={cn(
			"bg-primary text-primary-foreground animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 origin-(--bits-tooltip-content-transform-origin) z-50 w-fit text-balance rounded-md px-3 py-1.5 text-xs",
			className
		)}
		{...restProps}
	>
		{@render children?.()}
		<TooltipPrimitive.Arrow>
			{#snippet child({ props })}
				<div
					{...props}
					class={cn(
						"bg-primary top z-50 size-2.5 rotate-45 rounded-[2px]",
						side === "top" && "translate-x-1/2 translate-y-[calc(-50%_+_2px)]",
						side === "bottom" && "-translate-x-1/2 -translate-y-[calc(-50%_+_1px)]",
						side === "right" && "translate-x-[calc(50%_+_2px)] translate-y-1/2",
						side === "left" && "-translate-y-[calc(50%_-_3px)]"
					)}
				></div>
			{/snippet}
		</TooltipPrimitive.Arrow>
	</TooltipPrimitive.Content>
</TooltipPrimitive.Portal>
```

ieedan:
The resizable thing is because you (I assume) copied the shadcn styles but forgot to correct `-data-pane-group-direction` to the `data-direction` that pane forge uses. Just a simple find and replace 😊

ieedan:
Also we should probably mend the styling for tabs but idk if thats a priority for this PR 
CleanShot 2025-03-31 at 09 20 14

shyakadavis:
Screenshot 2025-03-31 at 22 29 54

🤔 Yeah, this is wrong.

shyakadavis:
> that's weird it wasn't doing that for me

I noticed that it only occurs when it's alone. When other tooltips are beside it, it's fine. 🤷

ieedan:
🤦‍♂️

Yeah what the heck man. It should be absolute so idk why that makes a difference...

ieedan:
https://github.com/user-attachments/assets/edaa204c-bb66-41d5-a89c-a91a5fad8985

Doesn't seem to be taking the styles...

ieedan:
I am so dumb but I found it. The `side` prop needs a default value. 
```diff
let {
	ref = $bindable(null),
	class: className,
	sideOffset = 0,
+	side = "top",
	children,
	...restProps
}: TooltipPrimitive.ContentProps = $props();
```

shyakadavis:
As for the tabs, I can push changes, but my question is, what look are we aiming for? The one on ui.shadcn.com, or the one v4.shadcn.com?

ieedan:
I think we just need to override the styles for the original. It should still look like v4 normally but for the component preview and stuff it should look like before.

huntabyte:
I think it's fine for now tbh @ieedan - we can worry about that before the final merge depending on what shad does with his docs!

huntabyte (approved):
You're awesome!

PR #1797, opened 2025-03-26
fix(v4): Generate correct code on `/theme`

ieedan:
This doesn't add the missing themes to `themes.css` but I think we are waiting on the original for those...

This consolidates the code for generating the themes so that we use the same methods for building the registry as we do for generating the code to copy.


ieedan:
@huntabyte not sure what the issue is but it seems like the themes page just refuses to work in preview. It can get the themes that we have configured to work locally though...

ieedan:
I set it up to build the other themes I still don't understand why some of them aren't showing up though...

huntabyte:
Nice looking theme 😂
CleanShot 2025-03-27 at 13 00 45@2x

huntabyte:
Yeah we'll see what shad ends up doing because right now it seems he's just supporting the gray variants on the v4 site.

ieedan:
> Nice looking theme 😂
> 
> CleanShot 2025-03-27 at 13 00 45@2x
> 
> 

I made it myself do you like it?

PR #1806, opened 2025-03-31
fix(v4): remove `cursor-pointer` where it doesn't belong

ieedan:
Removes `cursor-pointer` anywhere it's applied that's not a link I made some incorrect assumptions when changing some of these before.

Also adds some better styling indication to `/colors` to show that any part of the color card is clickable.


huntabyte (approved):
💯

PR #1807, opened 2025-03-31
fix(v4): Re-enable remote dependencies

ieedan:
Noticed this when reading through https://ui.shadcn.com/docs/registry/examples.

Some of the examples have urls in their registry dependencies. I broke support for this when #1781 got merged.

This PR re-adds support by expecting callers of `fetchRegistry` to supply a url instead of a name. It also modifies some of the other functions to conform with that expectation. This also allows you to pass registry items into `fetchRegistry` that don't belong to the same registry without both of them using the same base url.


packages/cli/src/utils/registry/index.ts
```diff
@@ -174,7 +174,8 @@ async function fetchRegistry(baseUrl: string, paths: string[]) {
 		return results;
 	} catch (e) {
 		if (e instanceof CLIError) throw e;
-		throw error(`Failed to fetch registry from ${baseUrl}. Error: ${e}`);
+
+		throw error(`Failed to fetch registry. Error: ${e}`);
```
ieedan, line 178: This looks like a regression but any errors should already be handled lower down and will display the url.

PR #1815, opened 2025-04-07
fix(docs)(next): z-index overlap in block toolbar on /blocks

cowboycodr:
The "View source" button was overlapping the header on the [/blocks](https://next.shadcn-svelte.com/blocks) route due to a `z-50` class. Removing it fixes the issue.

image


huntabyte (approved):
Thank you!

PR #1827, opened 2025-04-16
chore(v4): Docs fixes

### LINKED ISSUE #1823: Doc : Mode Watcher
### Change Type

Correction

### Proposed Changes

recently I 'm trying to start a new project using the CLI tools etc ..
then I want to install the mode watcher. but when copy the code in the docs, it's seems that the styling disappears somehow, 
```svelte
	
<script lang="ts">
  import { ModeWatcher } from "mode-watcher";
  let { children } = $props();
</script>
 
<ModeWatcher />
{@render children?.()}
```
the thing is that's it's should also add the import for the stylesheets : 
```svelte
	
<script lang="ts">
  import '../app.css';
  import { ModeWatcher } from "mode-watcher";
  let { children } = $props();
</script>
 
<ModeWatcher />
{@render children?.()}
```

ieedan:
Fixes #1823

- Fixes `<Steps/>` so that the numbers show up next to the headings again
- Removes `tailwind.config` section of docs since we don't use it anymore

### DIFF
--- a/sites/docs/src/content/components-json.md
+++ b/sites/docs/src/content/components-json.md
@@ -55,18 +55,6 @@ Configuration to help the CLI understand how Tailwind CSS is set up in your proj
 
 See the [installation section](/docs/installation) for how to set up Tailwind CSS.
 
-### tailwind.config
-
-Path to where your `tailwind.config.js` file is located.
-
-```json title="components.json"
-{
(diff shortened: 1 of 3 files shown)


huntabyte (approved):
Thank you!

PR #1840, opened 2025-04-21
feat(v4): Allow buttons with href to be disabled

ieedan:
Forgot to downstream this from `bits-ui` before and found myself needing it so I thought I would come and add it here.

bits-ui PR: https://github.com/huntabyte/bits-ui/pull/1055

### DIFF
--- a/sites/docs/src/lib/registry/ui/button/button.svelte
+++ b/sites/docs/src/lib/registry/ui/button/button.svelte
@@ -4,7 +4,7 @@
 	import { type VariantProps, tv } from "tailwind-variants";
 
 	export const buttonVariants = tv({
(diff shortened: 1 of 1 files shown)


huntabyte (approved):
You rock!

Wombosvideo:
May I ask what you use disabled links for?

huntabyte:
I could think of a few, maybe "locked" features where what appears to be a button takes you to another page in the app. Though validation to access the links page would happen server side, the client side touch is a nice one.

Wombosvideo:
I see. I am not really into teasing people but I can see the business side value for this. Thanks for clarification 👍🏼

ieedan:
I find myself using this a ton when there are features that are disabled for users that don't have a certain role / haven't paid for a certain feature that way if you have documentation that makes mention of a feature it's clear that the feature is just disabled for you and not missing entirely

Wombosvideo:
> I find myself using this a ton when there are features that are disabled [...] that way if you have documentation that makes mention of a feature it's clear that the feature is just disabled for you and not missing entirely

That actually makes a lot of sense! Thanks

PR #1848, opened 2025-04-22
fix: "set up" as verb

DePasqualeOrg:
"set up" is a verb, and "setup" is a noun.

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -21,7 +21,7 @@ jobs:
           cache: pnpm
 
       # Image cache setup
-      - name: Setup Image cache
+      - name: Set Up Image Cache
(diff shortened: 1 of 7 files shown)


huntabyte (approved):
Thank you! May be a while before we merge this as not in the mood to deal with a bunch of merge conflicts at the moment with the next and tailwind 4 branches!

AdrianGonz97:
I think we'll keep it the way it is in `main`. However, you feel up for it, feel free to make the changes to the `next` branch!

DePasqualeOrg:
@AdrianGonz97, I will create another PR for the `next` branch, but why not just merge this into `main`? LLMs consistently reproduce this mistake, especially in code comments and function names, because it's so prevalent in repos like this.

PR #1856, opened 2025-05-01
cli: custom registry & tailwind 4 stuff

huntabyte:

### DIFF
--- a/packages/cli/src/utils/registry/schema.ts
+++ b/packages/cli/src/utils/registry/schema.ts
@@ -1,13 +1,46 @@
 import * as v from "valibot";
 
 export type RegistryItemType = v.InferOutput<typeof registryItemTypeSchema>;
-const registryItemTypeSchema = v.picklist([
-	"registry:ui",
-	"registry:component",
-	"registry:example",
+
+const registryItemTargetType = ["registry:file", "registry:page"] as const;
+const registryItemSimpleType = [
+	"registry:lib",
 	"registry:block",
+	"registry:component",
+	"registry:ui",
 	"registry:hook",
-	"registry:page",
+	"registry:theme",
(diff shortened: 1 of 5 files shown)


packages/cli/src/utils/registry/schema.ts
```diff
@@ -13,14 +11,10 @@ const registryItemSimpleType = [
 	"registry:style",
 ] as const;
 
-/**
- * Used for internal purposes only.
- */
+/** Used for internal purposes only. */
```
huntabyte, line 14: gross

huntabyte: LMAO you can edit my comments!??

packages/cli/src/utils/updaters.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { CssVars } from "./registry/schema.js";
+
+export function updateCssVars(source: string, cssVars: CssVars): string {
+	if (Object.keys(cssVars).length === 0) return source;
+
+	const ast = parse(source);
+
+	// TODO: Consider adding status logs for each update. Perhaps this function could return a `string[]` of them.
+
+	// TODO: Should we error out if we fail to find/update any of the selectors?
```
huntabyte, line 11: I think it's better to be permissive here. Some projects won't have all selectors (like .dark) and it's fine to just skip updates we can't make rather than error out

huntabyte: eh but then again maybe failing silently isn't good either 🤔

huntabyte:
Doing some little unit tests before more comprehensive with fixtures

AdrianGonz97:
merging without an approval? smh

PR #1873, opened 2025-05-11
docs: Update vite.md

FaintShadow:
Solving the issue submmited by me #1831 by adding missing ES target (to the docs where it should have been specified) that leads to the sidebar component producing errors on build/dev.


AriDeve:
Should likely add _why_ the target is recommended to be set to `ES2022` or higher, essentially, add a note that includes the URL to [Svelte 5 typescript documentation](https://svelte.dev/docs/svelte/typescript).
Also, the lowest version allowed is `ES2015` even with `useDefineForClassFields` set to `true`

FaintShadow:
I will add those, thank you for mentioning it ❤️

PR #1877, opened 2025-05-12
fix: incorrect icon color on sidebar sub items

andersmmg:
The sidebar sub button icons were always colored the accent color, even when not active. This adds the active modifier.


huntabyte:
How we have it now is how it is in shadcn/ui which we treat as the source of truth.

Will close this for now but should it update there we can reopen here.

PR #1906, opened 2025-05-16
fix(tw4): Update calendar & range calendar styles

ieedan:
This just updates the sizes for now to match the original.


huntabyte (approved):
You're the man!

PR #1917, opened 2025-05-17
docs(twv4): Review registry docs

ieedan:
A few things I found going through the docs.

### DIFF
--- a/sites/docs/src/content/registry/index.md
+++ b/sites/docs/src/content/registry/index.md
@@ -15,7 +15,7 @@ out to us](https://github.com/huntabyte/shadcn-svelte/discussions).
 
 </Callout>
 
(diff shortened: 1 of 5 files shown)


sites/docs/src/content/registry/index.md
```diff
@@ -27,4 +27,4 @@ If you'd like to see an example of a registry, we have a [template project](http
 
 You can clone it using `degit`
 
-<PMExecute command="degit huntabyte/shadcn-svelte/registry-template" />
+<PMExecute command="degit huntabyte/shadcn-svelte/registry-template#next-tailwind-4" />
```
ieedan, line 30: Feel free to remove this if we merge straight to main. Or if we merge to next then we should just use `#next`

Issue #1949, opened 2025-05-19
next: Tabs margins misaligned


wiesson:
### Describe the bug

![Image](https://github.com/user-attachments/assets/9c50ecc2-52f6-4965-aba1-6b65758c4445)

### Reproduction

Compare https://next.shadcn-svelte.com/docs/components/tabs with https://ui.shadcn.com/docs/components/tabs 

### Logs

```bash

```

### System Info

```bash
System:
    OS: macOS 15.5
    CPU: (16) arm64 Apple M3 Max
    Memory: 89.14 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 11.4.0 - /usr/local/bin/npm
    bun: 1.2.13 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 136.0.7103.114
    Safari: 18.5
  npmPackages:
    bits-ui: ^1.5.1 => 1.5.1 
    formsnap: ^2.0.1 => 2.0.1 
    mode-watcher: ^1.0.7 => 1.0.7 
    svelte: ^5.30.2 => 5.30.2 
    svelte-sonner: ^1.0.1 => 1.0.1 
    sveltekit-superforms: ^2.25.0 => 2.25.0 
    tailwindcss: ^4.1.7 => 4.1.7 
    vaul-svelte: 1.0.0-next.7 => 1.0.0-next.7
```

### Severity

annoyance

PR #1950, opened 2025-05-19
fix: remove 1 pixel offset from tabs-trigger

MrVoshel:
Addresses #1949 

Before
image

After
image

### DIFF
--- a/sites/docs/src/lib/registry/ui/tabs/tabs-trigger.svelte
+++ b/sites/docs/src/lib/registry/ui/tabs/tabs-trigger.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
-	import { Tabs as TabsPrimitive } from "bits-ui";
 	import { cn } from "$lib/utils.js";
+	import { Tabs as TabsPrimitive } from "bits-ui";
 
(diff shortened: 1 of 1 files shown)


huntabyte:
Now that I'm looking, it seems like this is by design https://v4.shadcn.com ?

MrVoshel:
> Now that I'm looking, it seems like this is by design https://v4.shadcn.com ?

Yeah, looking at it and it's implemented in the same way.
After researching a bit it seems to be a subpixel rendering issue.

Screenshots from https://v4.shadcn.com (but same behaviour in the actual svelte version)
300% zoom - slight misalignment
image

400% zoom - aligned
image

Tried experimenting a bit with the margin too, but of course by using values such as 0.5px the same problems arise at some zoom levels.

Moving away from the, supposedly intended, original design seems the most straightforward approach of OCD-proofing it lol
but I'm assuming staying faithful to the original is preferred

huntabyte:
>  I'm assuming staying faithful to the original is preferred

Yeah if it is something intentional in the original it is preferred (which it looks to be given the specifics of those styles).

PR #1960, opened 2025-05-20
feat(next): make components more JSDoc friendly

ieedan:
From discord: https://discord.com/channels/1194404141389336576/1374023382424752248

This PR adds default values to any non-required props for components in the registry. The root of the problem on discord seems to be that JS is expecting a prop without a default value to be passed to the component. By providing a default value that expectation should go away.

### DIFF
--- a/sites/docs/src/lib/registry/ui/accordion/accordion-content.svelte
+++ b/sites/docs/src/lib/registry/ui/accordion/accordion-content.svelte
@@ -4,7 +4,7 @@
 
 	let {
 		ref = $bindable(null),
-		class: className,
+		class: className = '',
 		children,
 		...restProps
 	}: WithoutChild<AccordionPrimitive.ContentProps> = $props();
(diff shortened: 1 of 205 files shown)


ieedan:
I agree but I'm not quite sure how to approach it. Since the types get completely stripped out for the JS registry that's all that it really has to go off of.

huntabyte:
If anything we'd do this in the CLI after we strip the types vs in the main TS codebase

ieedan:
It would likely have to be before right? Because how else would we know to add default values to optional properties?

ieedan:
I am gonna attempt to do this in `strip-types` and see how it goes ideally we would just add `undefined` as a default value for optional properties.

AdrianGonz97:
This ended up being an error on the JS user's end. They had `lang="ts"` on their script tags when using the JS only version of our components.

PR #1966, opened 2025-05-20
chore: consolidate images

huntabyte:

### DIFF
--- a/sites/docs/src/lib/package-manager.ts
+++ b/sites/docs/src/lib/package-manager.ts
@@ -1,7 +1,7 @@
 import type { Cookies } from "@sveltejs/kit";
-import { type Agent, type Command, type ResolvedCommand } from "package-manager-detector";
-import { resolveCommand } from "package-manager-detector/commands";
 import { Context, PersistedState, watch } from "runed";
+import { resolveCommand } from "package-manager-detector/commands";
+import type { Agent, Command, ResolvedCommand } from "package-manager-detector";
 import { useCookie } from "./hooks/use-cookie.svelte.js";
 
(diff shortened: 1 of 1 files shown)


AdrianGonz97 (dismissed):

sites/docs/src/routes/(view)/view/[name]/+page.svelte
```diff
@@ -1,10 +1,15 @@
...
 
-	let { data }: { data: PageData } = $props();
+	let { data }: PageProps = $props();
+
+	const block = $derived(Blocks[data.block.name]);
 </script>
 
 <div class={cn("themes-wrapper bg-background", data.block?.container?.className)}>
-	<data.block.component />
+	{#await block.component() then BlockComponent}
```
Comment (huntabyte) [inline sites/docs/src/routes/(view)/view/[name]/+page.svelte:12]: This kinda sucks though right because we don't get these SSR'd?

Comment (huntabyte) [reply sites/docs/src/routes/(view)/view/[name]/+page.svelte:12]: Perhaps we can do the fetching of the block in the `+page.ts` using information from its `+page.server.ts`

Comment (AdrianGonz97) [reply sites/docs/src/routes/(view)/view/[name]/+page.svelte:12]: they were never ssr'd since svelte components cant be serialized. but yea, it does suck

AdrianGonz97 (approved):

sites/docs/src/routes/(view)/view/[name]/+page.svelte
Comment (AdrianGonz97) [reply sites/docs/src/routes/(view)/view/[name]/+page.svelte:12]: (especially when every instance of a block gets a new copy of the `Block` registry, which probably makes it slower to load

PR #2040, opened 2025-05-29
fix(next): Drawer padding consistency

ieedan:
This PR updates the drawer padding to be more consistent with the way that the dialog header and footer works.

It's worth noting that this is NOT how the original is styled. But to me it makes more sense this way.

CleanShot 2025-05-29 at 10 19 46


ieedan:
Yeah he is overriding it
CleanShot 2025-05-29 at 13 06 22


To me it seems a little silly to have to do this to your content so that it appears the right size relative to the header and footer every time.

PR #2043, opened 2025-05-30
chore: doc updates

huntabyte:


AdrianGonz97:
the animations for `sonner` appear to be broken (reduced motion is _not_ enabled) img

PR #2050, opened 2025-06-01
refactor!: remove `overflow-auto` from table wrapper

temich:
Using `overflow` breaks `position: sticky` for nested elements (like table header).
Having this in the table component is unintuitive and can lead to a lot of wasted time.


AdrianGonz97:

PR #2060, opened 2025-06-05
docs:Add Svelte5 repo link to the bug_report.yml.

rogerfar:
Adding a link to a repo Stackblitz for Svelte 5 using Shadcn next.

This would need a branch called "next" in the repo project: https://github.com/huntabyte/shadcn-repro-template


huntabyte:
Thank you! We'll be merging next with main _very_ soon so this won't be necessary 😁

PR #2065, opened 2025-06-06
chore: small tweaks

AdrianGonz97:


v4/src/lib/hooks/use-is-mac.svelte.ts
```diff
@@ -1,9 +1,5 @@
 export function useIsMac(): { readonly current: boolean } {
-	let isMac = $state(false);
-
-	$effect(() => {
-		isMac = navigator.platform.includes("MAC");
-	});
+	const isMac = $derived(navigator.platform.includes("MAC"));
```
huntabyte, line 2: Navigator isn't available on server

huntabyte (approved):

PR #2087, opened 2025-06-08
fix(cli): update failing on recent Node versions

WarningImHack3r:
`Dirent.path` is deprecated since Node 20.12.0, and has been replaced since then by `Dirent.parentPath`, which is essentially an alias.

`Dirent.path` seems to have been removed in Node 24, breaking for most recent Node versions in a similar way to #1003.

As shadcn-svelte officially only supports Node 22+, I think it's fair to replace the usage completely.

This is the only instance of `Dirent.path` I found throughout the entire codebase.

PR #2088, opened 2025-06-08
fix(cli): skip missing directories instead of throwing

WarningImHack3r:
Missing directories should be ignored rather than causing an error. I, for example, don't have any hooks, so no `hooks` directory. When I want to update my components, the CLI throws because it wants an empty directory instead of just ignoring it. This PR fixes that.

I didn’t encounter any issues after making this change, so it's safe to do as far as I’m aware.

~~Due to previous lint issues in `registry-template`, #2087 must be merged before this one.~~

### DIFF
--- a/.changeset/many-forks-train.md
+++ b/.changeset/many-forks-train.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix: don't throw on missing directories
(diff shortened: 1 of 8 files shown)


WarningImHack3r:
Ok, I don't know what's happening with the linting, but it seems like the Prettier rules inside registry-template are wrong compared to the root ones.
Unfortunately, the incorrect config takes over the correct one, which is not what we want.
The styling merged with my previous PR is correct; this one I did right now, running Prettier format from the repo root, should not be kept

AdrianGonz97 (approved):
thank you!

PR #2096, opened 2025-06-09
fix(docs): update component code viewer to use grid to prevent overflow

### LINKED ISSUE #2095: Component source overflow
### Change Type

Correction

### Proposed Changes

The new component source preview overflows and doesn't fit in the viewport.

Image

Wombosvideo:
Fixes the overflowing code issue in the component source code viewer.

Makes use of the CSS grid layout mode, which handles width more intuitively in some cases.

Fixes #2095

### DIFF
--- a/docs/src/lib/components/component-code-viewer/component-code-viewer-code.svelte
+++ b/docs/src/lib/components/component-code-viewer/component-code-viewer-code.svelte
@@ -9,14 +9,14 @@
 
 {#if file}
 	<div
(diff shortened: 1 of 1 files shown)


huntabyte:
This doesn't seem to fix the issue being described, which I believe they mean when the height is less than the minimum height it doesn't shrink. I'll add a fix for this later.

Wombosvideo:
It does however fix the x-Axis overflow a few lines have

huntabyte:
https://github.com/user-attachments/assets/cda4da55-036b-4770-8e32-59fd201dace2

X-Axis overflow is already working.

huntabyte:
Fixed in a recent PR, thank you anyways!

PR #2097, opened 2025-06-09
chore(cli): upgrade minimum supported Node to 20.12

WarningImHack3r:
As a follow-up to #2087, bump the minimum Node version to 20.12.x.


AdrianGonz97 (approved):
much appreciated!

PR #2133, opened 2025-06-13
fix: Only pad left side of sidebar submenu items for better UX

hhvrc:
The submenu items being padded on both sides makes the user click on an absent part of the sidebar when they want to select a menu subitem.
I think it would be better if subitems still are padded on the left side but extends all the way to the right.


huntabyte:
Thank you, but we're trying to closely track the original unless it makes 100% sense to deviate. If this gets merged into the upstream shadcn/ui then we can reopen this.

Issue #2038, opened 2025-05-29
Doughnut chart legend causes horizontal overflow when too many items are present


imlargo:
### Describe the bug

When rendering a <PieChart /> with the legend prop enabled and many data entries, the legend items overflow the horizontal bounds of the parent container (Card.Content), breaking the layout.

![Image](https://github.com/user-attachments/assets/160a045a-3d93-4c33-8b09-f9a912c52063)

### Reproduction

I’ve already included the following in my Tailwind content paths:
```
'./node_modules/svelte-ux/**/*.{svelte,js}',
'./node_modules/layerchart/**/*.{svelte,js}'
```

```

<script lang="ts">
	import * as Card from '$ui/card/index.js';
	import * as Chart from '$ui/chart/index.js';
	import { PieChart } from 'layerchart';

	const chartData = [
		{ browser: 'chrome', visitors: 320, color: '#2474f5' },
		{ browser: 'safari', visitors: 210, color: '#5996f7' },
		{ browser: 'firefox', visitors: 180, color: '#8fb8fa' },
		{ browser: 'edge', visitors: 145, color: '#bbd4fc' },
		{ browser: 'other', visitors: 95, color: '#e2ecfe' },
		{ browser: 'opera', visitors: 60, color: '#2474f5' },
		{ browser: 'brave', visitors: 75, color: '#2474f5' },
		{ browser: 'vivaldi', visitors: 40, color: '#2474f5' },
		{ browser: 'uc', visitors: 55, color: '#2474f5' },
		{ browser: 'maxthon', visitors: 30, color: '#2474f5' },
		{ browser: 'avant', visitors: 20, color: '#2474f5' },
		{ browser: 'konqueror', visitors: 15, color: '#2474f5' },
		{ browser: 'midori', visitors: 10, color: '#2474f5' },
		{ browser: 'netscape', visitors: 5, color: '#2474f5' },
		{ browser: 'seamonkey', visitors: 8, color: '#2474f5' }
	];

	const chartConfig = {
		visitors: { label: 'Visitors' },
		chrome: { label: 'chrome', color: '#2474f5' },
		safari: { label: 'safari', color: '#5996f7' },
		firefox: { label: 'firefox', color: '#8fb8fa' },
		edge: { label: 'edge', color: '#bbd4fc' },
		other: { label: 'other', color: '#e2ecfe' },
		opera: { label: 'opera', color: '#2474f5' },
[...]

PR #2139, opened 2025-06-14
fix: add chart context and legend components to fix chart legend horizontal overflow; refactor chart container

ptrtht:
## Problem
Chart legends were overflowing and overlapping with chart content when space was constrained. LayerChart's built-in legend positioning caused visual issues where legend items would bleed into the chart area or card footer, making charts unreadable in compact layouts. This directly references #2038 

## Solution
Implemented a temporary workaround using a flexible chart context system with external legend positioning until LayerChart addresses their legend overflow issues upstream.
### Temporary Components Added:
- ChartContext - Provides chart configuration to child components as a workaround
- ChartLegend - Auto-generates legend items from chart config (bypassing LayerChart's built-in legend)
- Updated ChartContainer - Works with or without context
### This approach allows us to:
- Immediately resolve legend overflow issues for users
- Maintain clean, type-safe APIs while waiting for upstream fixes
- Easily migrate back to LayerChart's native legend once their positioning is fixed
- Preserve backward compatibility during the transition
### Future Migration Path:
Once LayerChart resolves their legend positioning, we can deprecate `ChartLegend`, reverting to the simpler native `legend={true}` approach. 

## Key Features:
Auto-generation: Legends automatically derive items from chartConfig without manual data passing
Type Safety: Full TypeScript support with proper inference
Flexible Positioning: Legends can be positioned outside chart containers

### Existing usage (still works):
```svelte
<Chart.Container config={chartConfig}>
  <BarChart legend={true} /> <!-- LayerChart's native legend -->
</Chart.Container>
```
### New workaround for overflow issues:
```svelte
<Chart.Context config={chartConfig}>
  <Chart.Container>
    <!-- Chart content -->
  </Chart.Container>
  <Chart.Legend /> <!-- Positioned outside, no overflow -->
</Chart.Context>
```

### Alternative Approaches Considered
- CSS-only fixes - Attempted padding/margin adjustments, but LayerChart's absolute positioning made this unreliable across different chart types and screen sizes.
- Overriding LayerChart styles - Complex CSS overrides conflicted with LayerChart's internal layout calculations and broke responsiveness.

### DIFF
--- a/docs/registry.json
+++ b/docs/registry.json
@@ -390,6 +390,14 @@
 				},
 				{
 					"type": "registry:file",
+					"path": "src/lib/registry/ui/chart/chart-context.svelte"
+				},
+				{
+					"type": "registry:file",
+					"path": "src/lib/registry/ui/chart/chart-legend.svelte"
+				},
+				{
+					"type": "registry:file",
 					"path": "src/lib/registry/ui/chart/chart-style.svelte"
 				},
 				{
--- a/docs/src/lib/registry/blocks/chart-pie-legend.svelte
+++ b/docs/src/lib/registry/blocks/chart-pie-legend.svelte
@@ -28,29 +28,35 @@
 		<Card.Description>January - June 2024</Card.Description>
 	</Card.Header>
 	<Card.Content class="flex-1">
-		<Chart.Container config={chartConfig} class="mx-auto aspect-square max-h-[250px]">
-			<PieChart
-				data={chartData}
-				key="browser"
-				value="visitors"
-				label={(d) =>
-					d.browser
-						.split("")
-						.map((c, i) => (i === 0 ? c.toUpperCase() : c))
-						.join("")}
-				cRange={chartData.map((d) => d.color)}
-				props={{
-					pie: {
-						motion: "tween",
-					},
-				}}
-				legend
-			>
-				{#snippet tooltip()}
-					<Chart.Tooltip hideLabel />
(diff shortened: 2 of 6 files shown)


huntabyte:
Hey @ptrtht, really appreciate you taking the time to build this out, I can see a lot of though went into the workaround and keeping it clean.

That said, I'm hesistant to merge this as-is since it introduces a fair amount of complexity around a temporary problem. It's a solid stopgap, but long term I thinin this should be resolvable through LayerChart itself. I want to keep our API surface minimal and avoid maintaining parallel legend logic if we can help it.

Also, for a proper external legend solution, we really need LayerChart to expose something like a `LegendPayload` similar to `TooltipPayload` so we can build on solid internals rather than hacking something together if that makes sense.

ptrtht:
I agree this isn't ideal @huntabyte . Most pie charts have more than four legends, and simply ignoring the problem feels wrong, even if it's an upstream issue with layerchart.

Since [layerchart](https://github.com/techniq/layerchart/issues) doesn't seem to have a fix planned, my only other thought is to implement custom labels directly within our `ChartContainer`. The downside is this would restrict the label width to the chart's width, which I really dislike.

The original shadcn package supports this and the charts are awesome there, I would expect that we should also be supporting it in the svelte ecosystem.

How do you think we should proceed here? A temporary solution might be just creating a gist with the workaround without official support until the issue is fixed upstream.

PR #2149, opened 2025-06-19
fix: chat card font size

milksense:
Original [reference](https://github.com/shadcn-ui/ui/blob/84d6c83bad10e995bd648d59403f402ce7728444/apps/v4/components/cards/chat.tsx#L109)


huntabyte (approved):
Thank you!

PR #2150, opened 2025-06-19
Docs: Updated calendar component description

MeCandyCat:
Simple fix of description in the Calendar component docs page, where Shadcn Ui had a different term.


huntabyte:
Yeah I don't think his is correct in this case lol

huntabyte:
Thank you though!

PR #2163, opened 2025-06-24
fix: Use `--color-<color>` instead of `--<color>`

### LINKED ISSUE #2144: Sonner component is transparent with non-"latest" colors
### Describe the bug

https://github.com/huntabyte/shadcn-svelte/blob/d10b30b699eb3e705a7703e9ea8293fc18290af6/docs/src/lib/registry/ui/sonner/sonner.svelte#L11

This line causes the Sonner component to be rendered with a transparent background if the config is still not new with the `oklch` format, which was recently introduced.

For example, in my case, I have "old" color variables defined as raw HSL values, which are not usable as-is in the CSS without wrapping them inside `hsl()`.

For backward compatibility, the color used should be `--color-<mycolor>` instead of just `--<mycolor>`. Both syntaxes support this, and it doesn't break anything AFAICT.

What do you think about this little change?

### Reproduction

https://github.com/WarningImHack3r/svelte-changelog/tree/601ba9da3f04be06b1addb3a7efa84422a7d9864
(Start the dev server, head over to the homepage, and click a reaction inside any accordion item)

### Logs

```bash

```

### System Info

```bash
-
```

### Severity

annoyance

ieedan:
Fixes #2144

### DIFF
--- a/docs/src/lib/registry/ui/sonner/sonner.svelte
+++ b/docs/src/lib/registry/ui/sonner/sonner.svelte
@@ -8,6 +8,6 @@
 <Sonner
 	theme={mode.current}
 	class="toaster group"
-	style="--normal-bg: var(--popover); --normal-text: var(--popover-foreground); --normal-border: var(--border);"
+	style="--normal-bg: var(--color-popover); --normal-text: var(--color-popover-foreground); --normal-border: var(--color-border);"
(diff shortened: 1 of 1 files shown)


huntabyte (approved):
Thanks man!

PR #2164, opened 2025-06-24
fix(cli): publish `@shadcn-svelte/registry` and add `meta`

### LINKED ISSUE #2159: [registry]: missing keys in registrySchema, missing type Exports
### Prerequisites

- [x] This feature already exists in [shadcn/ui](https://github.com/shadcn/ui) - if not, it won't be considered here so don't continue with your issue.

### Describe the feature

First of off all, thank you for your amazing work.

As i'm trying to adapt the registry for my project, i found to missing things. In the ["original" implementation](https://github.com/huntabyte/shadcn-svelte/blob/main/packages/registry/src/schemas.ts) `registryItemSchema` is just using the `registryItemFileSchema` to use in `files`. In your implementation you missed for example `meta`. 

```json
		{
				"name": "comp-45",
				"type": "registry:component",
				"registryDependencies": ["label"],
				"files": [
					{
						"path": "src/lib/registry/default/components/comp-45.svelte",
						"type": "registry:component"
					}
				],
				"meta": { // will be missing in build
					"tags": ["input", "label"]
				}
			},
```

I don't want to open another issue for that, but it would be nice if you export the types and maybe the schema also from `shadcn-svelte`. 

```ts
import type { RegistryItem } from 'shadcn-svelte'; // Cannot find module 'shadcn-svelte' or its corresponding type declarations.ts(2307)
```

huntabyte:
Closes #2159

### DIFF
--- a/.changeset/huge-tigers-do.md
+++ b/.changeset/huge-tigers-do.md
@@ -1,0 +1,5 @@
+---
+"@shadcn-svelte/registry": patch
+---
+
+initial release
--- a/packages/cli/test/registry-template.test.ts
+++ b/packages/cli/test/registry-template.test.ts
@@ -38,6 +38,9 @@ it("should accurately build the registry template", async () => {
 		[
 		  {
 		    "description": "A simple hello world component",
+		    "meta": {
+		      "author": "huntabyte",
+		    },
 		    "name": "hello-world",
 		    "registryDependencies": [
 		      "button",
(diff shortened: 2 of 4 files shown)


AdrianGonz97 (approved):

max-got:
Hey, did you publish @shadcn-svelte/registry? Can' seem to install it 🙃

huntabyte:
oops, should be published now! @max-got sorry about that one!

Issue #2197, opened 2025-06-30
Navigation Menu docs use incorrect links in preview


MarvinZein:
### Change Type

Correction

### Proposed Changes

In the Home and Components entries of the Navigation Menu preview, some links link to non-existent routes.

For example, `docs/primitives/alert-dialog` should be `docs/components/alert-dialog`

PR #2198, opened 2025-06-30
docs: fix navigation menu links to use correct component routes

MarvinZein:
In the preview of the Navigation Menu several entries were pointing to non-existent or outdated routes. This PR updates those links so that each menu item now navigates to the actual route of its corresponding component.


MarvinZein:
Fixes #2197

huntabyte:
Thank you!

PR #2203, opened 2025-07-03
fix: `<Table.Row>` highlight no longer under the row borders

LyroyTheToad:
When hovering over a `<Table.Row>` the highlight would be seen under the row borders.

(Exaggerating border with `[&_tr]:border-b-8`)
Actual:
Actual

Expected:
Expected

To fix this, I had to move the highlight from the row to the cells.

The `... :[&,&>svelte-css-wrapper]: ...` selector is to handle the case where you define a CSS variable in a `<Table.Head>` or `<Table.Cell>`, like:
```svelte
<Table.Row>
  <Table.Cell --css-var={someReactiveVar}>
    Row
  </Tabel.Cell>
</Table.Row>
```
which would result in this HTML structure:
```html
<tr>
  <svelte-css-wrapper>
    <td>
      Row
    </td>
  </svelte-css-wrapper>
</tr>
```

### DIFF
--- a/docs/src/lib/registry/ui/table/table-cell.svelte
+++ b/docs/src/lib/registry/ui/table/table-cell.svelte
@@ -13,7 +13,10 @@
 <td
 	bind:this={ref}
 	data-slot="table-cell"
-	class={cn("whitespace-nowrap p-2 align-middle [&:has([role=checkbox])]:pr-0", className)}
+	class={cn(
+		"whitespace-nowrap bg-clip-padding p-2 align-middle [&:has([role=checkbox])]:pr-0",
+		className
+	)}
 	{...restProps}
 >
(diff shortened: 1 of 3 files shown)


LyroyTheToad:
@huntabyte sorry for the tag, one important consideration.

I've been playing around with tables and with defining css variables in table elements (e.g. `<Table.Cell --css-var={something}>`, and the `<svetle-css-wrapper>` is causing a lot of issues.
I had a hydration_mismatch warning from svelte, and some pages had a null to Object conversion error and failed to load (idk why).

I don't know if we should keep the `:[&,&>svelte-css-wrapper]:` selector, or if it should be removed to discurage defining css variable "svelte style" in tables.

PR #2208, opened 2025-07-05
fix(select): prevent content overflow with max-w-min

braden-w:
## Problem Description

The Select component had a significant UX issue where dropdown content with long text would overflow when constrained by the trigger's width. This created a poor user experience, especially in scenarios where:
- Select options contained descriptive text longer than the trigger width
- The trigger was intentionally kept narrow for UI consistency (e.g., `w-[120px]`)
- Options needed to display their full content for clarity

The issue manifested as horizontal text overflow within the dropdown, making longer options unreadable and breaking the component's usability.

## Solution Discovery Process

During the investigation, I experimented with several CSS approaches to solve this overflow issue:

1. **`w-max`** - Initially tried to let the content width grow to fit its content, but this didn't play well with the existing min-width constraint from `min-w-(--bits-select-anchor-width)`

2. **`max-w-none`** - Attempted to remove any maximum width constraints, but this could lead to unbounded growth and didn't provide the precise control needed

3. **`w-auto`** - Considered letting the browser automatically determine width, but this didn't resolve the fundamental constraint issue

4. **`max-w-min`** - The winning solution\! This sets `max-width: min-content`, which allows the dropdown to expand just enough to fit its longest content item while respecting the minimum width constraint

## Code Changes

The fix required a minimal, single-line addition to the Select.Content component:

```diff
// select-content.svelte
class={cn(
-  "bg-popover text-popover-foreground ... min-w-[8rem] overflow-y-auto overflow-x-hidden ...",
+  "bg-popover text-popover-foreground ... min-w-[8rem] max-w-min overflow-y-auto overflow-x-hidden ...",
  className
)}
```

## Minimal Reproduction Example

```svelte
<script lang="ts">
  import * as Select from "$lib/registry/ui/select/index.js";
  let value = $state("");
</script>

<Select.Root type="single" bind:value>
  <Select.Trigger class="w-[120px]">
    {value || "Pick"}
  </Select.Trigger>
  <Select.Content>
    <Select.Item value="short">Short option</Select.Item>
    <Select.Item value="long">This is a very long option that should overflow the narrow trigger</Select.Item>
    <Select.Item value="description">Option with description - Lorem ipsum dolor sit amet</Select.Item>
  </Select.Content>
</Select.Root>
```

## Benefits

1. **Improved Readability**: Users can now read complete option text without horizontal scrolling
2. **Maintained Design Consistency**: The dropdown still respects the minimum trigger width when options are short
3. **Zero Breaking Changes**: This is purely an enhancement - existing implementations continue to work exactly as before
4. **Minimal Code Impact**: A single CSS class addition solves the problem elegantly

## Demo

Before the fix:
```
┌─────────────┐
│ Pick     ▼  │ (Trigger: 120px wide)
└─────────────┘
       ↓
┌─────────────┐
│ This is a v…│ (Content truncated, overflow hidden)
└─────────────┘
```

After the fix:
```
┌─────────────┐
│ Pick     ▼  │ (Trigger: 120px wide)
└─────────────┘
       ↓
┌──────────────────────────────────────┐
│ This is a very long option that      │ (Content fully visible)
│ should overflow the narrow trigger   │
└──────────────────────────────────────┘
```

This fix ensures that shadcn-svelte's Select component provides an excellent user experience regardless of content length, matching the quality and attention to detail users expect from the library.

### DIFF
--- a/docs/src/lib/registry/ui/select/select-content.svelte
+++ b/docs/src/lib/registry/ui/select/select-content.svelte
@@ -22,7 +22,7 @@
 		{sideOffset}
 		data-slot="select-content"
 		class={cn(
-			"bg-popover text-popover-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 max-h-(--bits-select-content-available-height) origin-(--bits-select-content-transform-origin) relative z-50 min-w-[8rem] overflow-y-auto overflow-x-hidden rounded-md border shadow-md data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1",
+			"bg-popover text-popover-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 max-h-(--bits-select-content-available-height) origin-(--bits-select-content-transform-origin) relative z-50 min-w-[8rem] max-w-min overflow-y-auto overflow-x-hidden rounded-md border shadow-md data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1",
 			className
 		)}
 		{...restProps}


braden-w:
On second thought, I'm wondering if this should actually be the default behavior. While the visual consistency is nice, there are valid use cases where allowing the dropdown to expand beyond the trigger width improves usability:

1. **Readability**: When options have long descriptive text, horizontal expansion can be more readable than wrapping to multiple lines
2. **Scanning**: Users can more quickly scan options when they're displayed on a single line
3. **Accessibility**: Some users might find horizontally expanded content easier to read than wrapped text

Perhaps instead of making this the default, we could make it an optional behavior? Something like:

```svelte
<\!-- Default behavior: dropdown can expand -->
<Select.Content>
  ...
</Select.Content>

<\!-- Opt-in to constrained width -->
<Select.Content class="max-w-min">
  ...
</Select.Content>
```

Or even a prop:
```svelte
<Select.Content constrainWidth={true}>
  ...
</Select.Content>
```

This way users can choose the behavior that best fits their use case. What do you think?

huntabyte:
I think this should definitely be opt-in. Rather than the prop I think that class makes the most sense here otherwise someone could set `constrainWidth={true}` and then also set a custom `max-w-` which would basically cancel out what we've applied if that makes sense.

braden-w:
Yes, I totally agree. Would this be too much to include in the docs? This does seem be a strange fix (I never really use `max-w-min`), but it might be useful enough and weird enough of a solution that it might be worth including for others in the future.

braden-w:
I've opened #2432 to document the `max-w-min` pattern as an opt-in solution per our discussion. Closing this PR in favor of the docs-only approach.

PR #2210, opened 2025-07-06
fix: hook use-is-mac

milksense:


milksense:
@huntabyte :)

huntabyte:
Hm the way it is right now works for me for some reason? Maybe we should use a case insensitive regex?

milksense:
`navigator.platform` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform).

But alternatively `navigator.userAgentData` still not implemented in Safari `Version 26.0 (21622.1.16.11.2)`

Older browsers that don't support `navigator.userAgentData` still have a good fallback solution available.

```typescript
"userAgentData" in navigator
  ? navigator.userAgentData?.platform.toLowerCase().includes("mac") // or navigator.userAgentData?.platform.includes("macOS") but not stable value yet
  : navigator?.platform.toLowerCase().includes("mac") // or navigator?.platform.includes("Mac") stable but deprecated
```

milksense:
image_July-23

@huntabyte

PR #2220, opened 2025-07-12
fix: Use svelte:element instead of @html

ssebastianoo:
The `Chart` component was using this old syntax:
```tsx
 {@html `<style>${themeContents}</style>`}
 ```
 making the error `Unknown word themeContents` raise.

I fixed it by using the `<svelte:element this={'style'}>` component instead of the raw `@html`

Discord thread: https://discord.com/channels/1194404141389336576/1393230690245541949


Ashton-Sidhu:
This fixed the same issue I was having as well, thanks !

huntabyte (approved):
Thank you!

PR #2228, opened 2025-07-17
docs: added missing dropdown-menu component to installation notice

Le0X8:
well, it's just a tiny fix.

#### Why was that a problem?

Later on this page, the `dropdown-menu` component is used. Not installing it would lead to errors. As `table` and `data-table` are used on this page too, it would be nice to install everything at once.

### DIFF
--- a/docs/content/components/data-table.md
+++ b/docs/content/components/data-table.md
@@ -53,7 +53,7 @@ This guide will show you how to use [TanStack Table](https://tanstack.com/table)
 
 1. Add the `<Table />` component to your project along with the `data-table` helpers. These helpers enable TanStack Table v8 to work with Svelte 5 Snippets, Components, etc.
 
-<PMAddComp name="table data-table" />
+<PMAddComp name="table dropdown-menu data-table" />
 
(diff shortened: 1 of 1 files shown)


docs/content/components/data-table.md
```diff
@@ -53,7 +53,7 @@ This guide will show you how to use [TanStack Table](https://tanstack.com/table)
 
 1. Add the `<Table />` component to your project along with the `data-table` helpers. These helpers enable TanStack Table v8 to work with Svelte 5 Snippets, Components, etc.
 
-<PMAddComp name="table data-table" />
+<PMAddComp name="table dropdown-menu data-table" />
```
huntabyte, line 56: We should probably just put it right before the point we use the dropdown menu. It's a bit confusing to tell peeps to add table but then there's also a dropdown menu in there if that makes sense.

huntabyte (changes requested):

Le0X8:
Done :)

huntabyte (approved):
Awesome, thank you!

PR #2235, opened 2025-07-22
feat(drawer): enable drawer content scroll on overflow

braden-w:
This change enables proper vertical scrolling for drawer components when content exceeds the drawer height, inspired by @huntabyte 's [implementation in the `vaul-svelte` docs](https://stackblitz.com/edit/vaul-svelte-scrollable-with-inputs?file=src%2Froutes%2F%2Bpage.svelte). Previously, drawers with long content could overflow without proper scrolling behavior, making it difficult for users to access all content and resulting in poor mobile UX.

To accomplish this, I wrapped the `{@render children?.()}` in a `<div class="flex-1 overflow-y-auto">` container. The `flex-1` class ensures the content takes up remaining space after the drag handle at the top, while `overflow-y-auto` enables vertical scrolling.

### DIFF
--- a/docs/src/lib/registry/ui/drawer/drawer-content.svelte
+++ b/docs/src/lib/registry/ui/drawer/drawer-content.svelte
@@ -32,6 +32,8 @@
 		<div
 			class="bg-muted mx-auto mt-4 hidden h-2 w-[100px] shrink-0 rounded-full group-data-[vaul-drawer-direction=bottom]/drawer-content:block"
 		></div>
-		{@render children?.()}
+		<div class="flex-1 overflow-y-auto">
+			{@render children?.()}
+		</div>
 	</DrawerPrimitive.Content>
 </DrawerPrimitive.Portal>


huntabyte:
Does applying overflow-y-auto to the content not work as expected when you have overflowing content?

braden-w:
@huntabyte Unfortunately not, I tried also tried the `overflow-auto` code onto the `Drawer.Content` component, but it doesn't scroll properly, seems to overflow with a ton of random whitespace. You can look into the StackBlitz [here](https://stackblitz.com/edit/vaul-svelte-scrollable-with-inputs?file=src%2Froutes%2F%2Bpage.svelte) and try moving the `overflow-auto` class from the div into the `Drawer.Content`.

Putting `overflow-auto` on the inner div:

CleanShot 2025-07-24 at 16 48 34@2x

Putting it on the `Drawer.Content`. Notice the extra whitespace and how it's suddenly much taller.

CleanShot 2025-07-24 at 16 49 04@2x

CleanShot 2025-07-24 at 16 49 07@2x

Not sure if this is what's happening, but it seems like it's simultaneously adding a scroll bar because there's not enough space, but it also is taller, giving it more space than necessary compared to the previously computed amount of scrollthat would be needed. And so the scroll bar now has a bunch of extra white space to scroll into.

It's unclear to me if the change I propose is the best course of action, though.

1. Should we always wrap with the `div` and `overflow-y-auto`?
2. At the same time, maybe a code example in the docs would suffice so they know how to handle overflow?
3. Maybe this is something we should fix in `vaul-svelte`?

PR #2248, opened 2025-07-31
fix:Copy button in charts toolbar

### LINKED ISSUE #2258: Wrong code in clipboard button for charts
### Change Type

Correction

### Proposed Changes

Pressing on the `view code` button shows you the right code for that chart.
Pressing on the `Clipboard` button gives the code for the chart you first opened, the page with regardless of what chart you are currently on.

https://github.com/user-attachments/assets/2f5f6c79-7b64-4273-aa7c-8a32bed3913f

Prajwalg19:
As pointed out in https://github.com/huntabyte/shadcn-svelte/discussions/2244#discussioncomment-13935080, the copy button was not working as expected. 

The copy button only copies the code of the charts that we first land on. By default the charts page lands on `/charts/area`, so when a user navigates to `/charts/bar` and tries to copy any variant of the bar chart, it will still copy the code for area charts. 

I have fixed  it by changing the onMount directive to effect to track the change in chart data and update the code state accordingly.

closes #2258

### DIFF
--- a/docs/src/lib/components/chart-toolbar.svelte
+++ b/docs/src/lib/components/chart-toolbar.svelte
@@ -12,7 +12,7 @@
 	import ChartCopyButton from "./chart-copy-button.svelte";
 	import type { HTMLAttributes } from "svelte/elements";
 	import type { HighlightedBlock } from "../../routes/api/block/[block]/+server.js";
-
+	import { browser } from "$app/environment";
 	let {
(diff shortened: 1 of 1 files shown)


Prajwalg19:
@huntabyte :)

huntabyte:
I think this will not cleanup properly.

PR #2253, opened 2025-08-02
docs: fix resizable handle not clickable in block viewer

yeoularu:
### Fixed resizable handle becoming unclickable when dragged inward in docs block viewer

When the resizable handle was dragged to the left, it became unclickable due to being covered by a background element.

Added `z-20` to the resizable handle to keep it above other elements.


huntabyte (approved):
Thank you!

Issue #2219, opened 2025-07-11
Poor handling of RTL


Achraf112Ben:
### Describe the bug

Hello! 
While I was coding an app that targets mainly **RTL** (Right to Left) languages (Arabic, Hebrew, Farsi...) I've noticed that many shadcn-svelte components handle poorly the **RTL** orientation 
As an example: The close button on the Sheet component stays on the right side of the sheet even when the app is set to **RTL** and thus entangles with the sheet title 

<img width="796" height="458" alt="Image" src="https://github.com/user-attachments/assets/44db3860-6dfe-4d0b-98a1-d15c665c7f7f" />

Upon investigation I've found out that the problems stems from the usage of the properties **left** and **right** which are LTR iclusive in styling instead of **start** and **end** which adapt well to both **LTR** and **RTL**

<img width="805" height="215" alt="Image" src="https://github.com/user-attachments/assets/0f6b5cef-23b3-484b-a82c-47dd52057f63" />

I was able to fix the issue in my own code by replacing the **left** and **right** Tailwind CSS properties with **start** and **end** respectively 

<img width="805" height="215" alt="Image" src="https://github.com/user-attachments/assets/9a0bce8e-59e0-4a77-8e2f-bb081490fa54" />

<img width="797" height="335" alt="Image" src="https://github.com/user-attachments/assets/a721891b-44d0-4a25-9f6c-def8042ae6c9" />

I'm wondering if this could be fixed at the repository level so that **RTL** has a better support 

### Reproduction

<SheetPrimitive.Close
			class="ring-offset-background focus-visible:ring-ring rounded-xs focus-visible:outline-hidden absolute end-4 top-4 opacity-70 transition-opacity hover:opacity-100 focus-visible:ring-2 focus-visible:ring-offset-2 disabled:pointer-events-none"
		>

### Logs

```bash

```

### System Info

```bash
N/A
```

### Severity

annoyance

PR #2257, opened 2025-08-08
fix: properly handle rtl

omar-mohamed-khallaf:
Properly renders the close button to handle `rtl`.


Achraf112Ben:
Hello, 
I also approve this PR since I stumbled upon this issue myself about a month ago while developing an RTL Arabic app using Shadcn Svlete and opened up an issue #2219  about and proposed the same fix in this PR 
So, while looking in the PR I happened to find this PR that exactly solve the issue in #2219 
So I put my approve behind it.
Regards

huntabyte (approved):
Thank you!

PR #2259, opened 2025-08-11
docs: fix copying the correct source for current chart in chart-toolbar (#2258)

### LINKED ISSUE #2258: Wrong code in clipboard button for charts

CongHieuTruong:
## Description

Fixes #2258

This PR resolves a bug where the **"Copy"** button in the chart toolbar
always copied the code from the **first chart opened** on the page,
regardless of the chart currently being viewed.

### Steps to Reproduce
1. Open a chart page (e.g. *Area Chart – Interactive*).
2. Click **View Code** → correct code is displayed.
3. Click **Copy** → correct code is copied for this chart.
4. Navigate to a different chart (e.g. *Area Chart – Linear*).
5. Click **View Code** → correct code is displayed for the new chart.
6. Click **Copy** → **incorrect**: clipboard still contains code from the first chart.

### Root Cause
- Copy logic extracted highlighted HTML **only once** in `onMount` and never
  updated when the selected chart or file changed.
- Highlighted HTML contained encoded entities that required decoding before copying.

### Changes Made
- Replace `onMount`-only extraction with a reactive `$effect` that runs whenever
  the chart or file changes.
- Decode highlighted HTML on the client via a temporary `<pre>` element.
- Add SSR fallback by stripping HTML tags when DOM APIs are not available.

### Result
The **"Copy"** button now copies the correct raw source for the currently
displayed chart. No breaking changes.

### DIFF
--- a/docs/src/lib/components/chart-toolbar.svelte
+++ b/docs/src/lib/components/chart-toolbar.svelte
@@ -12,7 +12,6 @@
 	import ChartCopyButton from "./chart-copy-button.svelte";
 	import type { HTMLAttributes } from "svelte/elements";
 	import type { HighlightedBlock } from "../../routes/api/block/[block]/+server.js";
-	import { onMount } from "svelte";
 
 	let {
 		chart,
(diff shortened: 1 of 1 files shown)


CongHieuTruong:
@huntabyte looping back on PR #2259 for #2258 — tiny fix, zero risk, already tested.
Been sitting for a while now, and every day we delay, more users hit the bug.
This is an easy win we could ship today 🚀

huntabyte (approved):

PR #2262, opened 2025-08-13
fix: Reorder ALIASES const so that replacements run in a more sensical order

elliott-with-the-longest-name-on-github:
Since `lib` is typically a parent to all of the other aliases, it makes most sense to run it last -- otherwise the `lib` alias will clobber the `utils` alias and transform things to `$LIB$/utils.js` during registry build instead of transforming them to `$UTILS$`


AdrianGonz97:
thank you!

PR #2272, opened 2025-08-20
docs: Data-table return to render content for sorted list

HarveyBates:
Added required return statement to render a sorted list header. Otherwise the header will not display.


huntabyte:
it doesn't need a return statement there. There is no opening brackets for the function

HarveyBates:
🤦 thanks!

PR #2291, opened 2025-09-18
chore: Bump deps

ieedan:
Bumps some dependencies of the registry.

The only major is tailwind-variants but I think we are all good there from my testing and looking at the changelogs.


docs/package.json
```diff
@@ -58,28 +58,28 @@
...
-		"svelte": "^5.34.1",
+		"svelte": "^5.39.2",
 		"svelte-check": "^4.2.1",
 		"svelte-sonner": "^1.0.5",
 		"sveltekit-superforms": "^2.26.1",
 		"tailwind-merge": "^3.3.1",
-		"tailwind-variants": "^1.0.0",
-		"tailwindcss": "^4.1.10",
+		"tailwind-variants": "^3.1.1",
+		"tailwindcss": "^4.1.13",
```
huntabyte, line 79: `@tailwindcss/vite` version should align with this

PR #2296, opened 2025-09-22
chore: format tailwind classes correctly

abdel-17:
The tailwind classes in the `registry-template/` directory were not formatted correctly.


AdrianGonz97:
thanks for the pr! this should now be fixed as of #2825

Issue #2324, opened 2025-10-07
wrong usage example displayed on the new Empty component


HubbeDev:
### Change Type

Correction

### Proposed Changes

The usage example on [the new Empty component](https://www.shadcn-svelte.com/docs/components/empty) are wrong. Its displaying the the [Input group](https://www.shadcn-svelte.com/docs/components/input-group) usage example instead.

<img width="796" height="1573" alt="Image" src="https://github.com/user-attachments/assets/a050d114-59dd-4759-8b41-e3fef6e1db95" />

If no one got time to fix today. i can submit a pr later today.

PR #2328, opened 2025-10-07
docs: update empty component docs usage example.

HubbeDev:
Updates the correct usage example in the [new Empty ](https://www.shadcn-svelte.com/docs/components/empty#usage)component docs.

Closes: #2324

Edit: updated commit message.


huntabyte (approved):
Thank you!

PR #2401, opened 2025-11-11
fix(theme): add missing --destructive-foreground CSS variable

braden-w:
This PR adds the missing `--destructive-foreground` CSS variable to all Tailwind v4 theme definitions.

Currently, the `--destructive-foreground` variable is missing from the `:root`, `.dark`, and `@theme inline` blocks across template files and documentation. This causes the `text-destructive-foreground` utility to fall back to default foreground colors instead of providing proper high-contrast text (typically white) on destructive backgrounds.

Without this variable, destructive buttons and alerts don't have properly contrasting text colors, which can impact both aesthetics and accessibility.

---

The React version (shadcn/ui) already includes `--destructive-foreground` defined alongside `--destructive` in both light and dark modes in its templates:

- [templates/monorepo-next/packages/ui/src/styles/globals.css](https://github.com/shadcn-ui/ui/blob/main/templates/monorepo-next/packages/ui/src/styles/globals.css#L26)
- [apps/v4/content/docs/installation/manual.mdx](https://github.com/shadcn-ui/ui/blob/main/apps/v4/content/docs/installation/manual.mdx#L67)
- [packages/shadcn/test/fixtures/vite-with-tailwind/src/index.css](https://github.com/shadcn-ui/ui/blob/main/packages/shadcn/test/fixtures/vite-with-tailwind/src/index.css#L23)

Edit: notably, shadcn/ui's OKLCH templates have red-on-red instead of white-on-red `--destructive-foreground` values:

```css
  --destructive: oklch(0.577 0.245 27.325);
  --destructive-foreground: oklch(0.577 0.245 27.325);
```

I've filed https://github.com/shadcn-ui/ui/pull/8773 in shadcn/ui to fix this and confirm whether the red-on-red was intentional or should match their HSL templates (which correctly use white).

Either way, this PR still stands: we should still have a `--destructive-foreground` variable, just not sure if it should be white (matching `--foreground`) or red (matching `--destructive`).


braden-w:
Confirmed, shadcn/ui's OKLCH templates have red-on-red instead of white-on-red `--destructive-foreground` values.

I've filed [#8773](https://github.com/shadcn-ui/ui/pull/8773) in shadcn/ui to fix this and confirm whether the red-on-red was intentional or should match their HSL templates (which correctly use white).

Either way, the PR still stands: we should still have a `--destructive-foreground` variable, just not certain if it should be white (matching `--foreground`) or red (matching `--destructive`).

PR #2408, opened 2025-11-16
fix: make InputGroup direction-agnostic

skaldebane:
another rtl bug i just found, this time for `InputGroup`.

> sidenote: would an RTL toggle on the official samples help catch these types of issues?

### DIFF
--- a/docs/src/lib/registry/ui/input-group/input-group-addon.svelte
+++ b/docs/src/lib/registry/ui/input-group/input-group-addon.svelte
@@ -5,9 +5,9 @@
 		variants: {
 			align: {
 				"inline-start":
-					"order-first pl-3 has-[>button]:ml-[-0.45rem] has-[>kbd]:ml-[-0.35rem]",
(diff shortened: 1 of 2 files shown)


ieedan:
> another rtl bug i just found, this time for `InputGroup`.
> 
> > sidenote: would an RTL toggle on the official samples help catch these types of issues?

Curious if things would be broken in other ways because of that 🤔

Definitely something worth exploring though cause we get these issues quite a bit.

skaldebane:
> Curious if things would be broken in other ways because of that 🤔

Generally RTL shall work out-of-the-box (so if things break, that would be a bug), and it mostly already does, with few components here and there using absolute direction classes, and some configurable ones (e.g. sheet) only providing absolute sides as options (`left`, `right`).

Most of these remaining spots are pretty straightforward fixes, that shouldn't affect any existing LTR usages as far as I can see at least 😄

skaldebane:
I hope to tackle these issues as a whole as soon as I find time, going through every component and fixing any remaining issues in RTL layout direction.

PR #2417, opened 2025-11-21
fix: `Properties of undefined` error

ieedan:
Sorry for taking so long to do something about this everyone. It's unfortunately very difficult to reproduce because it manifests itself at seemingly random times. 

The fix here is to move all the components to their own files instead of re-exporting from bits-ui. The technical details of why this works is a mystery to me but it has fixed the error on every project I have tried it on. My best guess is that vite has trouble with the bits-ui import being in the index file and in each component individually.

### DIFF
--- a/docs/src/lib/registry/ui/alert-dialog/alert-dialog-content.svelte
+++ b/docs/src/lib/registry/ui/alert-dialog/alert-dialog-content.svelte
@@ -1,7 +1,9 @@
 <script lang="ts">
 	import { AlertDialog as AlertDialogPrimitive } from "bits-ui";
+	import AlertDialogPortal from "./alert-dialog-portal.svelte";
 	import AlertDialogOverlay from "./alert-dialog-overlay.svelte";
 	import { cn, type WithoutChild, type WithoutChildrenOrChild } from "$lib/utils.js";
+	import type { ComponentProps } from "svelte";
 
 	let {
 		ref = $bindable(null),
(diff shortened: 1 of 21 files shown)


docs/src/lib/registry/ui/alert-dialog/alert-dialog-portal.svelte
huntabyte: I think we'd want to use these instead of Bits UI directly in the components that use them, like `alert-dialog-content` which uses Portal etc!

saturnonearth:
Sorry to comment on this when it's already closed but I wanted to ask - what error exactly was this producing?

ieedan:
For whatever reason (and not consistently) bits-ui primitives would be undefined in the index file causing you to get "Error reading properties of undefined reading Root"

saturnonearth:
> For whatever reason (and not consistently) bits-ui primitives would be undefined in the index file causing you to get "Error reading properties of undefined reading Root"

Ah I see okay - I don't think I've had that issue. Just to double check (again), the error did not display as `ReferenceError: __SVELTEKIT_PAYLOAD__ is not defined` correct?

ieedan:
Nope haven't seen that one before

rudiv:
Is this the issue that's meant to fix [this discussion thread](https://github.com/huntabyte/shadcn-svelte/discussions/2252)?

If so, we updated components from the registry for `Select`, and it still doesn't work with the latest Bits UI.

ieedan:
If you update your components this should fix it

PR #2420, opened 2025-11-21
fix: RTL issues

ieedan:
Not a permanent fix we still need proactive about this. A lint rule would be preferable but it doesn't seem like one exists at this time.


mustafa0x:
Awesome work thanks so much @ieedan

I came to report an RTL issue but found this fixed it!

PR #2431, opened 2025-11-29
fix: add RTL support for dialog content

AnsBdran:
Fix the `DialogContent` component position on `rtl` markup direction


AnsBdran:
will comeback soon...

Issue #2439, opened 2025-12-02
DropdownMenu: CheckboxGroup unusable without passing value prop


Blackwidow-sudo:
### Describe the bug

Currently its not possible to use the `DropdownMenu.CheckboxGroup` without passing the `value` prop.
You get the following error:
```
Uncaught Svelte error: props_invalid_value
Cannot do `bind:value={undefined}` when `value` has a fallback value
https://svelte.dev/e/props_invalid_value
```

### Reproduction

https://stackblitz.com/edit/github-2cqe9hcw?file=src%2Froutes%2F%2Bpage.svelte

### Logs

```bash
Uncaught Svelte error: props_invalid_value
Cannot do `bind:value={undefined}` when `value` has a fallback value
https://svelte.dev/e/props_invalid_value

	in <unknown>
	in dropdown-menu-checkbox-group.svelte
	in menu-group.svelte
	in dropdown-menu-group.svelte
	in text-selection-layer.svelte
	in dismissible-layer.svelte
	in escape-layer.svelte
	in focus-scope.svelte
	in floating-layer-content.svelte
	in popper-content.svelte
	in popper-layer-inner.svelte
	in popper-layer.svelte
	in dropdown-menu-content.svelte
	in portal-consumer.svelte
	in undefined
	in portal.svelte
	in dropdown-menu-portal.svelte
	in dropdown-menu-content.svelte
	in floating-layer.svelte
	in menu.svelte
	in dropdown-menu.svelte
	in +page.svelte
	in +layout.svelte
	in root.svelte

    at props_invalid_value (chunk-FHMHV7DR.js?v=7fc9da63:405:19)
    at prop (chunk-YCDIHY23.js?v=7fc9da63:3445:18)
    at Menu_checkbox_group (bits-ui.js?v=7fc9da63:22993:104)
    at chunk-YCDIHY23.js?v=7fc9da63:296:58
    at update_reaction (chunk-FHMHV7DR.js?v=7fc9da63:3820:18)
    at update_effect (chunk-FHMHV7DR.js?v=7fc9da63:3948:21)
    at create_effect (chunk-FHMHV7DR.js?v=7fc9da63:3303:7)
    at branch (chunk-FHMHV7DR.js?v=7fc9da63:3479:10)
    at chunk-YCDIHY23.js?v=7fc9da63:290:17
    at update_reaction (chunk-FHMHV7DR.js?v=7fc9da63:3820:18)
[...]

PR #2440, opened 2025-12-02
fix(DropdownMenu): provide fallback value for checkbox-group to prevent props_invalid_value error

Blackwidow-sudo:
Fixes https://github.com/huntabyte/shadcn-svelte/issues/2439

Currently its not possible to use `DropdownMenu.CheckboxGroup` without passing the `value` prop.

### DIFF
--- a/registry-template/src/lib/registry/ui/button/button.svelte
+++ b/registry-template/src/lib/registry/ui/button/button.svelte
@@ -2,7 +2,7 @@
 	import { tv, type VariantProps } from "tailwind-variants";
 
 	export const buttonVariants = tv({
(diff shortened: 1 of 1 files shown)


Blackwidow-sudo:
> Thank you!

sry seems like some unintended commits slipped in 😂

ieedan:
Just gonna link this as reference for what the prop should be in these cases https://github.com/huntabyte/bits-ui/blob/798d69505806943dfa0d8c0d83e9f1ac0ca86e4d/packages/bits-ui/src/lib/bits/menu/components/menu-checkbox-group.svelte#L15

PR #2443, opened 2025-12-04
docs: fix NativeSelect import path to components

wiscaksono:
The Native Select docs currently import from $lib/registry/ui/native-select/index.js, but the component source and other examples use $lib/components/ui/native-select/index.js. This PR aligns the docs with the correct path for consistency and to prevent copy‑paste errors.

- Scope: docs/content/components/native-select.md
- Change: 1 addition, 1 deletion matching the Native Select page examples on shadcn‑svelte
- Motivation: Keep “Component Source” and “Usage” sections consistent so users can import the right module


huntabyte (approved):
Thank you!

PR #2451, opened 2025-12-09
chore: update funding

huntabyte:


braden-w:
Woo! Love seeing @ieedan added here. Been following his work on shadcn-svelte-extras and jsrepo for a while; huge fan 😊 Well deserved and an awesome crossover!

PR #2469, opened 2025-12-14
Fix: svelte reference

jdgamble555:
With the new svelte updates, the [state_referenced_locally](https://svelte.dev/docs/svelte/compiler-warnings#state_referenced_locally) warning is more pervasive.  This ensures it passes `pnpm check`.

There is also a warning:
```
The field "pnpm.onlyBuiltDependencies" was found in C:\web projects\shadcn-svelte\registry-template/package.json. 
This will not take effect. You should configure "pnpm.onlyBuiltDependencies" at the root of the workspace instead.
```
This is pointless when not in the root, so it was removed.

J

### DIFF
--- a/docs/src/lib/registry/ui/carousel/carousel-next.svelte
+++ b/docs/src/lib/registry/ui/carousel/carousel-next.svelte
@@ -3,7 +3,7 @@
 	import type { WithoutChildren } from "bits-ui";
 	import { getEmblaContext } from "./context.js";
 	import { cn } from "$lib/utils.js";
-	import { Button, type Props } from "$lib/components/ui/button/index.js";
+	import { Button, type Props } from "$lib/registry/ui/button/index.js";
 
 	let {
 		ref = $bindable(null),
--- a/docs/src/lib/registry/ui/carousel/carousel-previous.svelte
+++ b/docs/src/lib/registry/ui/carousel/carousel-previous.svelte
@@ -3,7 +3,7 @@
 	import type { WithoutChildren } from "bits-ui";
 	import { getEmblaContext } from "./context.js";
 	import { cn } from "$lib/utils.js";
-	import { Button, type Props } from "$lib/components/ui/button/index.js";
+	import { Button, type Props } from "$lib/registry/ui/button/index.js";
 
 	let {
 		ref = $bindable(null),
--- a/docs/src/lib/registry/ui/toggle-group/toggle-group-item.svelte
+++ b/docs/src/lib/registry/ui/toggle-group/toggle-group-item.svelte
@@ -2,7 +2,7 @@
 	import { ToggleGroup as ToggleGroupPrimitive } from "bits-ui";
 	import { getToggleGroupCtx } from "./toggle-group.svelte";
 	import { cn } from "$lib/utils.js";
-	import { type ToggleVariants, toggleVariants } from "$lib/components/ui/toggle/index.js";
+	import { type ToggleVariants, toggleVariants } from "$lib/registry/ui/toggle/index.js";
 
 	let {
 		ref = $bindable(null),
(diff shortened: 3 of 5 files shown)


huntabyte:
> This is pointless when not in the root, so it was removed.

This was here because the idea is people can clone that specific directory and they would then want that there, so the error can be ignored here in this repo :)

Thank you!

huntabyte:
Also I see you got the carousel to shut up about the warnings, but that feels like a bug from Svelte and we shouldn't need to hack around it like this tbh

jdgamble555:
WDYM? The docs suggest writing context like that to pass reactive variables. This is now the recommended way.

https://svelte.dev/docs/svelte/compiler-warnings#state_referenced_locally

They recommend hacking it that way.

huntabyte:
> They recommend hacking it that way.

Their example doesn't use a proxied object so in that example it makes sense. In our usage the object itself can cross function boundaries and retain reactivity so the warning is lying/useless here.

jdgamble555:
Hmmm... I wonder if you should do a warning ignore, and if you should run `pnpm check` in GH Actions?

jdgamble555:
@huntabyte - It looks like you're running the [check](https://github.com/huntabyte/shadcn-svelte/blob/9a7f140a6fd7fa78876f00551d87a6973433f019/.github/workflows/ci.yml#L26), but if you allow the warnings, you're forcing all your users to deal with the same warnings when they run check as well.

```bash
c:\web projects\post-talk\src\lib\components\ui\carousel\carousel.svelte:25:3
Warn: This reference only captures the initial value of `orientation`. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally (svelte)
                scrollNext,
                orientation,
                canScrollNext: false,

c:\web projects\post-talk\src\lib\components\ui\carousel\carousel.svelte:29:12
Warn: This reference only captures the initial value of `opts`. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally (svelte)
                handleKeyDown,
                options: opts,
                plugins,

c:\web projects\post-talk\src\lib\components\ui\carousel\carousel.svelte:30:3
Warn: This reference only captures the initial value of `plugins`. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally (svelte)
                options: opts,
                plugins,
                onInit,

c:\web projects\post-talk\src\lib\components\ui\chart\chart-container.svelte:20:27
Warn: This reference only captures the initial value of `id`. Did you mean to reference it inside a closure instead?
https://svelte.dev/e/state_referenced_locally (svelte)

        const chartId = `chart-${id || uid.replace(/:/g, '')}`;

c:\web projects\post-talk\src\lib\components\ui\toggle-group\toggle-group.svelte:36:3
Warn: This reference only captures the initial value of `variant`. Did you mean to reference it inside a closure instead?
https://svelte.dev/e/state_referenced_locally (svelte)
        setToggleGroupCtx({
                variant,
                size,

c:\web projects\post-talk\src\lib\components\ui\toggle-group\toggle-group.svelte:37:3
Warn: This reference only captures the initial value of `size`. Did you mean to reference it inside a closure instead?
https://svelte.dev/e/state_referenced_locally (svelte)
                variant,
                size,
                spacing

c:\web projects\post-talk\src\lib\components\ui\toggle-group\toggle-group.svelte:38:3
Warn: This reference only captures the initial value of `spacing`. Did you mean to reference it inside a closure instead?
https://svelte.dev/e/state_referenced_locally (svelte)
                size,
                spacing
        });

====================================
svelte-check found 0 errors and 7 warnings in 3 files
```

There are also now issues with eslint, `errors`, not just warnings...

```bash
C:\web projects\post-talk\src\lib\components\ui\button\button.svelte
  63:8  error  Unexpected href link without resolve()  svelte/no-navigation-without-resolve

C:\web projects\post-talk\src\lib\components\ui\chart\chart-tooltip.svelte
  9:40  error  '_payload' is defined but never used  @typescript-eslint/no-unused-vars

✖ 2 problems (2 errors, 0 warnings)
```

I'm happy to do a PR to at least ignore the `state_referenced_locally` errors, and fix the other two, but I don't want to waste my time if you don't approve it.

J

PR #2479, opened 2025-12-30
docs: Include instructions for adding to an existing project

mrgnw:
The instructions currently have a command for adding tailwindcss to a new Svelte project, but not for an existing project.

This adds the command to add tailwindcss to an existing project: `sv add tailwindcss`

### DIFF
--- a/docs/content/installation/sveltekit.md
+++ b/docs/content/installation/sveltekit.md
@@ -14,7 +14,7 @@ description: How to setup shadcn-svelte in a SvelteKit project.
 
 <Steps>
 
-### Add tailwind to your project
+### Add TailwindCSS to your project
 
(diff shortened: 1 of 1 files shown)


ieedan:
One of the Svelte CLI maintainers actually just put up a PR to change this to the way it is now 😅 https://github.com/huntabyte/shadcn-svelte/pull/2458 I don't have a lot of strong opinions on this but I would prefer not to ping-pong between solutions

mrgnw:
What are the odds they just made that change!

I think these two PR's aren't necessarily opposing each other —

I'd say #2458 is about streamlining the process for new projects, since the sv cli can now both create the project and add tailwind.

My hope with this PR would be to make the steps for adding to an *existing* project clear (using `sv add`).

PR #2487, opened 2026-01-15
fix: use getters in toggle-group context to avoid Svelte 5 reactivity warning

n00ki:
Fixes a Svelte 5 warning in the **ToggleGroup** component:

When props are destructured and passed directly to setContext, Svelte 5 warns because the values are captured at initialization time rather than being read reactively.

**The fix**: 
Convert direct prop values to getter functions, ensuring values are read lazily when accessed. This matches the pattern already used in `chart-container.svelte`.

**Before:**
```js
setToggleGroupCtx({
    variant,
    size,
    spacing,
});
```

**After:**
```js
setToggleGroupCtx({
    get variant() { return variant; },
    get size() { return size; },
    get spacing() { return spacing; },
});
```

Related: https://svelte.dev/e/state_referenced_locally

### DIFF
--- a/docs/src/lib/registry/ui/toggle-group/toggle-group.svelte
+++ b/docs/src/lib/registry/ui/toggle-group/toggle-group.svelte
@@ -33,9 +33,15 @@
 	}: ToggleGroupPrimitive.RootProps & ToggleVariants & { spacing?: number } = $props();
 
 	setToggleGroupCtx({
-		variant,
-		size,
-		spacing,
+		get variant() {
+			return variant;
+		},
+		get size() {
+			return size;
+		},
+		get spacing() {
+			return spacing;
+		},
 	});
 </script>
 


ieedan:
This does fix the problem but it seems like unnecessary overhead just to fix a warning that (a warning that doesn't provide much value in this case)

In these cases I think the right solution is probably just to suppress the warning with a comment as annoying as that might be...

n00ki:
@ieedan Thanks for the feedback! You're right that in typical usage these props are static. 
We can add `eslint-disable-next-line state_referenced_locally` to the function or `svelte-ignore state_referenced_locally` for each prop, yet this won't maintain full reactivity or be consistent with `chart-container.svelte` which uses the same pattern.

ieedan:
chart-container is setup that way because we are preparing for the user to change those props after the component has been mounted. In this case we shouldn't need the reactivity since those props are unlikely to change.

PR #2497, opened 2026-01-20
chore: Improve type-safety for rendering components

Stadly:
This PR aims to improve type-safety when providing props to components, at the exense of more verbose error messages in some cases. The most important case is when no props are passed to a component expecting props. **The old version gives no type-safety in that case.** Here are the resulting changes in type-safety error messages.

## No props component

```svelte
<!-- no-prop-component.svelte -->
no props
```

### No props provided
No change.
image

### Unexpected prop provied
Less verbose error message in old version when props are provided but not expected.
image
image

## Single prop component

```svelte
<!-- single-prop-component.svelte -->
<script lang="ts">
  type Props = { value: number };
  let { value }: Props = $props();
</script>

{value}
```

### No props provided
**Type-safety in new version when props are required but not provided.**

image
image

### Empty props provided
Less verbose error message in old version when required prop is not provided.

image
image

### Expected props provided
No change.

image

### Prop has wrong type
Less verbose error message in old version when prop has wrong type.

image
image

### Only unexpected prop provided
Better error message in new version when extra props are provided, but required prop is not provided.

image
image

### Unexpected prop provided
No change.

image

### DIFF
--- a/docs/src/lib/components/cards/payments.svelte
+++ b/docs/src/lib/components/cards/payments.svelte
@@ -80,13 +80,13 @@
 				renderComponent(Checkbox, {
 					checked: table.getIsAllPageRowsSelected(),
 					indeterminate: table.getIsSomeRowsSelected() && !table.getIsAllRowsSelected(),
-					onCheckedChange: (v) => table.toggleAllPageRowsSelected(!!v),
+					onCheckedChange: (v: boolean) => table.toggleAllPageRowsSelected(v),
 					"aria-label": "Select all",
 				}),
 			cell: ({ row }) =>
 				renderComponent(Checkbox, {
 					checked: row.getIsSelected(),
-					onCheckedChange: (v) => row.toggleSelected(!!v),
+					onCheckedChange: (v: boolean) => row.toggleSelected(v),
 				}),
 			enableSorting: false,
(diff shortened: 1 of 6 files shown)


ieedan:
I could be misunderstanding but in most cases this seems to move the error message from where it probably should be (on the props) to the component. 

This is probably backwards
```ts
export class RenderComponentConfig<
	TProps extends Record<string, unknown>,
	TComponent extends Component<TProps>,
> {
```

We want to infer the props from the component so that the error messages make sense.

Stadly:
@ieedan Yes, I completely agree. If there is a way to make it work properly when inferring the props from the component, that would be perfect! Unfortunately, I have not been able to find a solution where it works. When inferring the props from the component, one easily runs into false positive type errors.

Here is an example project where the current implementation of `renderComponent` (incorrectly) emits type errors, while the one from this PR does not: https://www.sveltelab.dev/jsa01s4t4ixc15v?files=.%2Fsrc%2Flib%2Frender-helper.ts

image

Stadly:
Turs out the changes in the PR exposed a few type issues in the code base concerning `renderComponent` (function args being `any`) and `renderSnippet` (missing params). I've updated the PR with fixes.

PR #2503, opened 2026-01-24
fix: add execCommand fallback for clipboard api in safari

### LINKED ISSUE #2502: "Copy Page" button does not work on Safari.
Same as issue on bits-ui docs https://github.com/huntabyte/bits-ui/issues/1878. Copy page button does not work on Safari. Gives no interaction, just fails.

addisonking:
resolves #2502; referenced https://github.com/huntabyte/bits-ui/pull/1880.

add execCommand fallback to clipboard hook for safari's transient activation restrictions.


ieedan:
Actually just fixed this in extras where this original came from so we can just use the same logic here :)

Issue #2484, opened 2026-01-06
createSvelteTable has poor performance on larger tables


Majora320:
### Describe the bug

The `createSvelteTable` wrapper around Tanstack Table is provided as part of the data-table package. It currently manages state by wrapping the table options in a new proxy layer every time state changes. This results in increasingly deep callstacks and performance that worsens the longer a page is left open, especially for larger tables. Propose using a single proxy instead that references a mutable `$state` variable.

### Reproduction

https://stackblitz.com/edit/github-6rqofbhy-fshvnzyw?file=src%2Froutes%2F%2Bpage.svelte

### Logs

```bash

```

### System Info

```bash
not relevant
```

### Severity

annoyance

jay-dee7:
Using tanstack table has been a pain for my project too, switched to regular shadcn-svelte table instead. Works much better

valentinschabschneider:
We noticed massive performance degradation after upgrading svelte.

old working version: 5.38.10
after upgrading: 5.46.4

We didn't investigate which version specifically introduced the issue, but it seems to be a problem with tanstack table as mentioned.

jay-dee7:
Interesting, I've had similar issues with layerchart too and degrading to a similar old version made it all better.

kenowi-dev:
I have the same problem and created a PR with a potential solution and some more details on my findings [PR 2511](https://github.com/huntabyte/shadcn-svelte/pull/2511).

I implemented the TanStack Table column resizing feature and this problem occures very fast, since resizing a column makes a lot of state changes.

The issue seems to be that on each state change the table options are updated with a proxy that merges the old and new values. 
**But** TanStack Table is also merging the old and new values. Since the new value that TanStack receives is a proxy this proxy gets merged into a second proxy. This repeats and increases on each state update.

## Potential solution
Change the `updateOptions()` function insude the `src/lib/components/ui/data-table/data-table.svelte.ts` file to the following:
```Js
function updateOptions() {
    table.setOptions(() => {
      return mergeObjects(resolvedOptions, options, {
        state: mergeObjects(state, options.state || {}),
        onStateChange: (updater: Updater<TableState>) => {
          if (updater instanceof Function) state = updater(state);
          else state = mergeObjects(state, updater);

          options.onStateChange?.(updater);
        },
      });
    });
  }
```
This removes the merging of the previous state so that `table.setOptions()` always receives a "new" object and the previous state is not proxied twice.

[...]

PR #2511, opened 2026-02-01
fix(data-table): mergeObjects causing nested proxies on each state change

kenowi-dev:
This PR fixes the nested proxy property access inside the data table component that leads to performance degradation the more state changes happened.

### Related 
- [Discussion 2456](https://github.com/huntabyte/shadcn-svelte/discussions/2456)
- [Issue 2484](https://github.com/huntabyte/shadcn-svelte/issues/2484)

## The Issue
The `mergeOptions` function is merging objects by using a proxy to retain the svelte reactivity of the underlying data.
This `mergeOptions` function is passed to the TanStack Table on initial creation.
On each state change an effect runs that updates the table options by merging the previous options with the new one, including the state.
```Js
function updateOptions() {    
    table.setOptions((prev) => {
      return mergeObjects(prev, options, {
        state: mergeObjects(state, options.state || {}),
        [...]
      });
    });
  }
  updateOptions();
  $effect.pre(() => updateOptions());
```

The Problem occurs since TanStack Table **also** calls the `mergeOptions` function on each `table.setOptions()` call with the previous options and the new ones. Since the new options is a merged proxy object it will be nested again.
```Js
// Pseudo code
// 1. First Proxy Creation
mergeObjects(prev, options, {state: ...}) 
/* creates */ new Proxy(prevTableOptions + newTableOptions)
// 2. Nested Proxy creation
table.setOptions(new Proxy(prevTableOptions + newTableOptions) 
/* creates */ -> new Proxy(previousTableOptions + new Proxy(prevTableOptions + newTableOptions)
```

Since the proxies are lazily merged, accessing a property causes the `get` and `findSourceWithKey` function to run for every proxy until the underlying data is retrieved. This leads to a massive slowdown especially when a lot of state changes happened.

image

## Solution
This PR removes the merging of the previous state inside the `updateOptions()` function. 
```Js
 function updateOptions() {
    table.setOptions(() => {
      return mergeObjects(resolvedOptions, options, {
        state: mergeObjects(state, options.state || {}),
        [...]
      });
    });
  }
```
With this approach the `table.setOptions()` always receives a new proxy object that does not rely on any "old" object and keeps the depth while accessing a propery constant.

## Additional Toughts
It is a bit difficult to wrap my head around all the proxy creation. Since `table.setOptions()` is still using the previous state, which is a proxy object, there might still be some nesting happening. But since the `findSourceWithKey` function inside `mergeObjects` is accessing the sources from newest to oldest, the nestin has no effect.  
Maybe a solution without any proxies at all would be nicer, but I was unable to find any.

I noticed the problem while trying to implement the column resizing feature of the TanStack Table. Since resizing creates a lot of state changes the problem is prominent very quickly.

### DIFF
--- a/docs/src/lib/registry/ui/field/field-error.svelte
+++ b/docs/src/lib/registry/ui/field/field-error.svelte
@@ -19,7 +19,7 @@
 		if (children) return true;
 
 		// no errors
-		if (!errors) return false;
+		if (!errors || errors.length === 0) return false;
 
 		// has an error but no message
 		if (errors.length === 1 && !errors[0]?.message) {


ieedan:
Thank you!

Issue #2482, opened 2026-01-02
A bug with Dialog + Form/Superform when `Submit` action is not working as expected.


0rangeFox:
### Describe the bug

Hey folks 👋, I am following the logic of creating dialogs with the form through the [link](https://shadcn-svelte.com/docs/components/dialog), but apparently something is wrong. In both examples in the [link](https://shadcn-svelte.com/docs/components/dialog) and in the code below, whenever I click on the `Dialog. Trigger` button, it triggers the `Submit` action of the forms and not the button inside the dialog. I have already tried dragging the `Dialog.Trigger` button outside the `form` element, and even so, the submit button inside the forms is never submitted.

Below is some sample code that is ready and quick to execute. If you look at the server-side console, you will see that none of the `console.log` are ever called when you click the `Submit` button with the `Dialog. Trigger` outside the `form` element, but if `Dialog.Trigger` is already inside the `form` element, you will notice that it is always called and the messages from `console.log` will appear in the console. Is there any explanation or reason why this happens?

### Reproduction

`Schema` type:
```typescript
    import { z } from 'zod';

    const testSchema = z.object({
        nickname: z.string().nonempty('Nickname is required!')
    });
```

`+page.server.ts`:
```typescript
import type { Actions, PageServerLoad } from './$types';
import type { RequestEvent } from '@sveltejs/kit';
import { superValidate } from 'sveltekit-superforms';
import { zod4 } from 'sveltekit-superforms/adapters';

async function onPageServerLoad() {
    return {
        form: await superValidate(zod4(testSchema))
    };
}

async function onRegister(event: RequestEvent): Promise<any> {
    console.log(event);
    const form = await superValidate(event, zod4(testSchema));
    console.log(form);
}

export const load: PageServerLoad = onPageServerLoad;
export const actions: Actions = {
    default: onRegister
};
```

`+page.svelte`:
```svelte
<script lang='ts'>
    import SuperDebug, { defaults, superForm } from 'sveltekit-superforms';
    import { zod4, zod4Client } from 'sveltekit-superforms/adapters';
    import * as Dialog from '$lib/components/ui/dialog';
    import * as Form from '$lib/components/ui/form';
    import { Input } from '$lib/components/ui/input';
    import { buttonVariants } from '$lib/components/ui/button';

    const form = superForm(defaults(zod4(testSchema)), {
        validators: zod4Client(testSchema)
    });

    const { form: formData, enhance } = form;
[...]

PR #2539, opened 2026-02-23
docs: Update dialog demo to fix auto submit issue 

MustafaHi:
The demo provide a code that has issue with `form`'s, which trigger submit when opening a dialog
modified the code to move the `form` element inside `dialog.content` to fix it
also added a note in the code

a reference to issue : https://github.com/huntabyte/shadcn-svelte/issues/2482

### DIFF
--- a/docs/src/lib/registry/examples/dialog-demo.svelte
+++ b/docs/src/lib/registry/examples/dialog-demo.svelte
@@ -6,9 +6,9 @@
 </script>
 
 <Dialog.Root>
-	<form>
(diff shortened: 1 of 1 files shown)


ieedan:
I think this isn't quite the right way to fix this since this is changing the meaning of the markup. It's reasonable to allow users to wrap a dialog in a form since a dialog could be the way a user selects the value of a field so moving the form within the dialog isn't always the correct solution.

For now I will merge #2476 but I think by default going forward we are going to add `type="button"` to the dialog trigger and other buttons.

MustafaHi:
Right, having type button by default is the correct fix

PR #2554, opened 2026-03-11
Use hydratable for random width in sidebar menu

Stadly:
This prevents the skeleton width from changing during hydration.


Stadly:
`hydratable` was introduced in Svelte version 5.44.0 (November 2025). Do we need to support older versions of Svelte?

PR #2558, opened 2026-03-12
chore: use `const` to comply with `svelte/prefer-const` lint rule

filiabel:
I think this is a nice default to avoid linting issues, and I assume many users have `svelte/prefer-const` rule enabled.
Feel free to close if it is deemed irrelevant!

**Side question:** It seems to have been a while since somebody ran `pnpm format`? I get 272 changed files. Any reason for that?


huntabyte:
> Side question: It seems to have been a while since somebody ran pnpm format? I get 272 changed files. Any reason for that?

It could be a new rule that was added that we haven't formatted against yet.

PR #2561, opened 2026-03-13
Fix WithElementRef type incompatibility with Svelte 5 bind:this

cellulosa:
This PR fixes a TypeScript incompatibility between WithElementRef and Svelte 5’s bind:this typing when used with certain DOM elements such as `<select>`.

The current implementation types ref as HTMLElement, which causes svelte-check to report an error when binding to more specific element types like HTMLSelectElement.

**Example error:**

Type 'HTMLSelectElement' is not assignable to type 'HTMLElement'. The types returned by 'remove()' are incompatible between these types. Type 'void' is not assignable to type 'Element'.

**Cause**

Svelte’s svelte/elements DOM typings differ slightly from the standard lib.dom.d.ts. In particular, remove() is typed differently, which makes HTMLSelectElement incompatible with HTMLElement in strict TypeScript contexts.

Because WithElementRef defines:

```
type WithElementRef<T> = T & {
  ref?: HTMLElement | null;
};
```

this forces bind:this={ref} to expect HTMLElement, which conflicts with elements like HTMLSelectElement.

**Fix**

Instead of forcing HTMLElement, make the element type generic so the ref matches the actual DOM element being bound.

```
type WithElementRef<Props, El extends Element> = Props & {
  ref?: El | null;
};
```

### DIFF
--- a/docs/src/lib/registry/ui/native-select/native-select.svelte
+++ b/docs/src/lib/registry/ui/native-select/native-select.svelte
@@ -1,7 +1,13 @@
 <script lang="ts">
-	import { cn, type WithElementRef } from "$lib/utils.js";
+	import { cn } from "@ts-packages/shadcn/generated/lib/utils.js";
 	import type { HTMLSelectAttributes } from "svelte/elements";
 	import ChevronDownIcon from "@lucide/svelte/icons/chevron-down";
+
+	type Props = HTMLSelectAttributes & {
+		ref?: HTMLSelectElement | null;
+		value?: string;
+		children?: any;
+	};
 
 	let {
 		ref = $bindable(null),
@@ -9,7 +15,7 @@
 		class: className,
 		children,
 		...restProps
-	}: WithElementRef<HTMLSelectAttributes> = $props();
+	}: Props = $props();
 </script>
 
 <div
@@ -24,7 +30,7 @@
 			"border-input placeholder:text-muted-foreground selection:bg-primary selection:text-primary-foreground dark:bg-input/30 dark:hover:bg-input/50 h-9 w-full min-w-0 appearance-none rounded-md border bg-transparent px-3 py-2 pe-9 text-sm shadow-xs transition-[color,box-shadow] outline-none disabled:pointer-events-none disabled:cursor-not-allowed",
 			"focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring-[3px]",
 			"aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive",
-			className
+			className,
 		)}
 		{...restProps}
 	>


ieedan:
I think the error here is valid unfortunately it seems like the code is just slop.

This could be a 1 line fix if we just change `HTMLSelectAttributes` -> `HTMLSelectElement`

cellulosa:
mh but then I get all sorts of errors:

```
	let {
		ref = $bindable(null),
		value = $bindable(),
		class: className,
		children,
		...restProps
	}: WithElementRef<HTMLSelectElement> = $props();
```

Property 'class' does not exist on type 'HTMLSelectElement & { ref?: HTMLElement | null | undefined; }'.

```
	<select
		bind:value
		bind:this={ref}
		data-slot="native-select"
		class={cn(
			"border-input placeholder:text-muted-foreground selection:bg-primary selection:text-primary-foreground dark:bg-input/30 dark:hover:bg-input/50 h-9 w-full min-w-0 appearance-none rounded-md border bg-transparent px-3 py-2 pe-9 text-sm shadow-xs transition-[color,box-shadow] outline-none disabled:pointer-events-none disabled:cursor-not-allowed",
			"focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring-[3px]",
			"aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive",
			className
		)}
		{...restProps}
	>
```

Property 'class' does not exist on type 'HTMLSelectElement & { ref?: HTMLElement | null | undefined; }'.

Argument of type '{ autocomplete: AutoFill; disabled: boolean; form: HTMLFormElement | null; labels: NodeListOf<HTMLLabelElement>; length: number; multiple: boolean; name: string; options: HTMLOptionsCollection; ... 339 more ...; "bind:value": string; }' is not assignable to parameter of type 'HTMLProps<"select", HTMLAttributes<any>>'.
  Type '{ autocomplete: AutoFill; disabled: boolean; form: HTMLFormElement | null; labels: NodeListOf<HTMLLabelElement>; length: number; multiple: boolean; name: string; options: HTMLOptionsCollection; ... 339 more ...; "bind:value": string; }' is not assignable to type 'Omit<HTMLSelectAttributes, never>'.
    Types of property 'form' are incompatible.
      Type 'HTMLFormElement | null' is not assignable to type 'string | null | undefined'.
        Type 'HTMLFormElement' is not assignable to type 'string'.

`{@render children?.()}`

This expression is not callable.
  Type 'HTMLCollection' has no call signatures.

ieedan:
Yeah I think the correct type should be:
```ts
WithElementRef<HTMLSelectAttributes,HTMLSelectElement>
```

cellulosa:
That also seems to error

Error: Type 'HTMLSelectElement' does not satisfy the constraint 'HTMLElement'.
  The types returned by 'remove()' are incompatible between these types.
    Type 'void' is not assignable to type 'Element'. (ts)
		...restProps
	}: WithElementRef<HTMLSelectAttributes, HTMLSelectElement> = $props();
</script>

ieedan:
So testing right now I actually don't get an error at all when using it like this:

```svelte
<script lang="ts">
    import * as NativeSelect from "$lib/registry/ui/native-select/index.js";

    let ref = $state<HTMLSelectElement | null>(null);
</script>

<NativeSelect.Root bind:ref>
    <NativeSelect.Option value="1">1</NativeSelect.Option>
    <NativeSelect.Option value="2">2</NativeSelect.Option>
    <NativeSelect.Option value="3">3</NativeSelect.Option>
</NativeSelect.Root>
```

cellulosa:
I'll try and see if I can reproduce the error on a fresh repo

Issue #2512, opened 2026-02-02
Right-to-left support for shadcn/ui components and CLI to migrate to RTL


Prajwalg19:
### Prerequisites

- [x] This feature already exists in [shadcn/ui](https://github.com/shadcn/ui) - if not, it won't be considered here so don't continue with your issue.

### Describe the feature

RTL Tab in the documentation with missing RTL features to be included in shadcn-svelte

ieedan:
Linking https://x.com/shadcn/status/2017287507881164816?s=20

itskazemk:
Can't wait

PR #2578, opened 2026-03-21
feat: add localization-aware layout support

Alja7dali:
## Summary

First, thank you for the remarkable work on `shadcn-svelte`.

This PR improves localization-aware layout support in the docs app and shared registry-backed UI primitives.

The goal is to make RTL/LTR behavior correct by default for newly generated registry code, while preserving existing LTR behavior. This is not a new component or a library scope expansion; it is an implementation improvement for existing surfaces so bilingual applications have a more reliable foundation.

## Problem

The current implementation still hardcodes physical `left` / `right` assumptions in several shared paths:

- root document direction is static
- shared primitives expose physical placement behavior
- directional icons are not mirrored consistently
- high-use docs/create surfaces rely on physical spacing and alignment utilities

This makes RTL support fragile and causes newly installed registry components to inherit layout behavior that is harder to localize.

## Changes

- add a localization contract with normalized `languageTag` and `direction`
- persist localization state in user config and keep the document root in sync
- apply `lang` and `dir` on SSR and client boot
- add shared helpers for resolving logical inline sides (`start` / `end`) at component boundaries
- update shared primitives such as `Sheet`, `Sidebar`, submenu content wrappers, pagination, carousel, calendar, range-calendar, and breadcrumbs to behave correctly in RTL/LTR
- replace physical inline utilities with logical ones in high-use docs/create components where reading flow matters
- mirror directional icons intentionally instead of relying on implicit physical semantics
- set explicit default `dir="ltr"` in `repro` and `registry-template`

## Why this belongs here

I understand that `shadcn-svelte` is a port of `shadcn/ui` and that the project does not accept net-new components or unrelated feature additions unless they exist upstream.

This PR stays within that boundary:

- no new components
- no new design patterns
- no i18n framework integration
- no scope expansion beyond layout correctness

It only improves how existing components and generated code behave in localized layouts, which helps teams building bilingual products without requiring downstream patches for the same RTL/LTR issues.

## Impact

After this change:

- the docs app has an explicit direction contract
- newly generated registry code is more localization-safe by default
- consuming apps still need to set a real document-level `dir` / `lang` for RTL behavior to activate
- previously installed component copies are not automatically updated


ieedan:
Hey @Alja7dali thank you the work but there is already an upstream implementation of this that (while I don't necessarily agree with) we intend to copy in the future see #2512 

In that implementation the CLI will do most of the work for handling this with some additional classes on some registry components as well.

Once that lands feel free to put up a PR for any things we missed because there will be quite a lot of surface area to cover!

Alja7dali:
Hey @ieedan thanks for the clarification and for linking #2512!

That makes sense — I’ll hold off and follow the upstream/parity approach instead.

Happy to contribute later once it lands to help cover any missed RTL/localization cases.

Appreciate it 🙌

PR #2533, opened 2026-02-19
docs updates

maxffarrell:
## Summary

Complete overhaul of the shadcn-svelte documentation to align with the current component API, new components, and design system updates.

### Docs changes

- **All 58 component docs rewritten** — updated API tables, usage examples, and code snippets for every component (accordion, alert, avatar, badge, breadcrumb, button, calendar, card, carousel, chart, checkbox, collapsible, combobox, command, context-menu, data-table, date-picker, dialog, drawer, dropdown-menu, field, hover-card, input, input-group, input-otp, item, kbd, label, menubar, native-select, navigation-menu, pagination, popover, progress, radio-group, range-calendar, resizable, scroll-area, select, separator, sheet, sidebar, skeleton, slider, sonner, spinner, switch, table, tabs, textarea, toggle, toggle-group, tooltip, typography, and more)
- **Changelog restructured** — split into per-release files under `docs/content/changelog/`
- **New components documented** — `Empty`, `Item`, `InputGroup`, `Kbd`, `ButtonGroup`, `NativeSelect`, `Spinner`, `Field`

### Component / example changes

- New and updated example files across the registry (`docs/src/lib/registry/examples/`)
- Resizable: added `orientation` prop alias for `direction`; fixed type to make `direction` optional with `$derived` reactivity
- Avatar: added `group`, `badge`, and `group-count` sub-components
- Custom shadcn-svelte prop documentation: `showCloseButton`, `variant="line"` on Tabs, `AlertDialog.Media`, `size` on Switch/AlertDialog

### Infrastructure

- Updated `pnpm-workspace.yaml`, `package.json`, `pnpm-lock.yaml` for dependency updates
- `velite.config.js` added for content processing
- `tsconfig.json`, `vite.config.ts`, `svelte.config.js` updated
- CLI preset support (`packages/cli/src/preset/`)
- Registry schema updates (`packages/registry/src/schemas.ts`)

## Test plan

- [ ] `pnpm check` passes with 0 errors and 0 warnings
- [ ] `pnpm lint` passes
- [ ] Component docs pages render correctly in the docs site
- [ ] Component preview examples render without errors
- [ ] Resizable component works with both `direction` and `orientation` props

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2595, opened 2026-03-25
fix: package manager executable command for bun

ahm0xc:


ieedan:
We generally use package-manager-detector for this so we expect their implementation to be correct. If it isn't this should probably be a change made upstream.

But I am pretty sure this is correct since they have been around a while now.

maxffarrell:
I noticed this as well and updated it in #2533 but going to reexamine

PR #2602, opened 2026-03-27
fix: add missing fields in schema of components.json

CatBraaain:
This PR fixes a mismatch between the `components.json` generated by `bunx shadcn-svelte@latest init` and the published schema definition of shadcn-svelte.

## Problem

`bunx shadcn-svelte@latest init` generates `components.json` which includes the following fields:

<details>
<summary>Example</summary>

```json
{
  "$schema": "https://shadcn-svelte.com/schema.json",
  "tailwind": {
    "css": "src/style/global.css",
    "baseColor": "neutral"
  },
  "aliases": {
    "components": "$lib/components",
    "utils": "$lib/utils",
    "ui": "$lib/components/ui",
    "hooks": "$lib/hooks",
    "lib": "$lib"
  },
  "typescript": true,
  "registry": "https://shadcn-svelte.com/registry",
  "style": "vega",
  "iconLibrary": "lucide",
  "menuColor": "default",
  "menuAccent": "subtle"
}
```
</details>

- `iconLibrary`
- `menuColor`
- `menuAccent`
 
However, these fields are not defined in the current schema.json, resulting in validation errors such as:

```
Property iconLibrary is not allowed.
Property menuColor is not allowed.
Property menuAccent is not allowed.
```

## Changes

Added the following missing fields to schema.json:

* `iconLibrary`
* `menuColor`
* `menuAccent`

This ensures consistency between the CLI output and the schema, eliminating unnecessary validation errors.

### DIFF
--- a/.changeset/lovely-tigers-glow.md
+++ b/.changeset/lovely-tigers-glow.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix: add missing fields to components.json schema
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -86,6 +86,35 @@
 					"additionalProperties": false
 				}
 			]
+		},
+		"iconLibrary": {
+			"description": "The icon library to use for your components.",
+			"type": "string",
+			"enum": [
+				"lucide",
+				"tabler",
+				"hugeicons",
+				"phosphor",
+				"remixicon"
+			]
+		},
+		"menuColor": {
+			"description": "The color scheme for menus.",
+			"type": "string",
+			"enum": [
+				"default",
+				"inverted",
+				"default-translucent",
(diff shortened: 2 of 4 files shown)


CatBraaain:
This change only affects the documentation.
Since the docs package is versioned as `0.0.0` and not released, a changeset is not required, I think.

ieedan:
> This change only affects the documentation. Since the docs package is versioned as `0.0.0` and not released, a changeset is not required, I think.

It changes the CLI code so we need a changeset for the CLI

CatBraaain:
hi @ieedan
Thanks for reviewing so quick.
I added changeset and using ICON_LIBRARIES instead of PRESET_ICON_LIBRARIES.
Please check this when you have a moment.

PR #2662, opened 2026-04-17
fix: epicenter sponsor logo visibility on light backgrounds

braden-w:
The Epicenter sponsor card renders its logo as two overlapping circles—gray and white—with no background. On dark mode this looks fine, but on light mode the white circle is invisible against the light card surface. The logo appears as a single gray crescent instead of the actual mark.

The fix adds a black `<rect>` background to the inline SVG, matching the canonical Epicenter logo (which is always two circles on a black background—the black is part of the identity). The wrapper div gets `overflow-hidden` + `rounded-lg`/`rounded-md` to clip the square rect into a rounded mark shape:

```svelte
<!-- before: circles float with no background -->
<svg viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">
  <circle cx="170" cy="170" r="100" fill="#cccccc" opacity="1" />
  <circle cx="230" cy="230" r="100" fill="#ffffff" opacity="1" />
</svg>

<!-- after: proper mark with black background -->
<div class="size-10 overflow-hidden rounded-lg ...">
  <svg viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">
    <rect width="400" height="400" fill="#000" />
    <circle cx="170" cy="170" r="100" fill="#ccc" />
    <circle cx="230" cy="230" r="100" fill="#fff" />
  </svg>
</div>
```

Also fixes the tagline from "Open source, local first apps" to "Local-first, open source apps"—hyphenated compound adjective, and leads with the differentiator instead of table-stakes "open source."

### DIFF
--- a/docs/src/lib/components/cta-mobile.svelte
+++ b/docs/src/lib/components/cta-mobile.svelte
@@ -10,17 +10,18 @@
 	></div>
 
 	<div
-		class="relative z-10 size-8 shrink-0 opacity-80 transition-opacity group-hover:opacity-100"
+		class="relative z-10 size-8 shrink-0 overflow-hidden rounded-md opacity-80 transition-opacity group-hover:opacity-100"
 	>
 		<svg viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">
-			<circle cx="170" cy="170" r="100" fill="#cccccc" opacity="1" />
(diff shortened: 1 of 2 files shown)


ieedan (approved):
2x Thanks

PR #2672, opened 2026-04-21
create docs: improve search

maxffarrell:
## Summary

shadcn/ui uses Fumadocs which is React only, used Flexsearch to best emulate

- Adds `docs/scripts/build-search-data.ts` to pre-build a FlexSearch index at build time
- Adds `docs/src/lib/utils/search.ts` for client-side search utilities
- Adds `/api/search.json` API endpoint
- Adds `dev:search` and `build:search` scripts to `docs/package.json`
- Adds `flexsearch` and `remove-markdown` as dependencies


maxffarrell:
I don't think the svecodocs implementation can be used cleanly here, this is broken up per section vs per page

ieedan:
Hey @maxffarrell finally getting around to taking a look at this.

It doesn't seem to be working for me as I would expect it to

https://github.com/user-attachments/assets/cfbf036d-a6af-4684-8889-b531bf057e35

PR #2674, opened 2026-04-21
create docs: create layout tweaks & mobile optimizations

maxffarrell:
## Summary

image
- mobile sponsor banner on /create and fits card below sidebar on desktop
- mobile optimized picker and paste preset responsive dialog
- changing theme resets to new theme's defaults, matching shadcn/ui behavior
- Removes `initialize-dialog.svelte` and `+page.server.ts` (replaced by client-side approach)
- Adds `ResetDialogCtx` export to design system provider
- Updates `preview/[item]` route to match shadcn/ui
- updates initialize dialog


packages/cli/src/preset/preset.ts
maxffarrell, line 1: this is the part I'm least sure of, but it seems like shadcn did it this way

PR #2678, opened 2026-04-21
create docs: update non-component docs, changelog, and velite config

maxffarrell:
## Summary

- Adds `/docs/changelog` route with RSS feed (page.svelte + page.ts)
- Adds changelog entries for Luma, Sera, component-composition, plus all historical changelog entries broken out into their own pages
- Updates CLI, Forms, Installation, and Theming docs
- Adds Skills docs
- Adds changelog collection to `velite.config.js` (new `date` field on docSchema, new `changelog` collection)
- Adds `getChangelogPages` and `ChangelogPage` exports to `src/lib/docs.ts`


ieedan:
CleanShot 2026-05-14 at 13 45 07
CleanShot 2026-05-14 at 13 45 23

Just comparing ours to theirs on this PR. We are missing the skills page from the sidebar and maybe we can update our groups/ordering to be the same or it can be a separate PR.

ieedan:
CleanShot 2026-05-14 at 13 47 56

The sera preview SS is also missing is that just because we don't have a reference SS yet?

docs/content/cli.md
```diff
@@ -105,15 +105,3 @@ Options:
   -o, --output <path>  destination directory for json files (default: ./static/r)
   -h, --help           display help for command
 ```
-
```
ieedan, line 108: This is still a thing we don't need to remove it

docs/content/skills.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+## How It Works
+
+1. **Project detection** — The skill activates when it finds a `components.json` file in your project.
+2. **Context injection** — It reads your project configuration and injects it into the assistant's context.
+3. **Pattern enforcement** — The assistant follows shadcn-svelte composition rules: using `Field.Group` for forms, namespace imports, Svelte 5 runes, and correct Bits UI APIs.
+4. **Component discovery** — The assistant checks installed components and local docs before generating code.
+
+<!--## Differences from shadcn/ui
```
ieedan, line 65: is this just claude or are we keeping this for a different reason?

docs/content/installation/index.md
```diff
@@ -5,10 +5,42 @@ description: How to install dependencies and structure your app.
...
-## Guides
+<Callout class="mb-6 border-emerald-600 bg-emerald-100 dark:border-emerald-400 dark:bg-emerald-900">
+
+**Recommended for new projects:** Use [shadcn-svelte/create](/create) to build your preset visually and generate the right setup command for your framework.
+
+</Callout>
+
+Choose the setup that matches your starting point.
+
+<SetupCards />
```
ieedan, line 23: I'm tempted to say to just leave the existing docs intact for this because this only adds confusion. We won't have complete parity but while we aren't doing any scaffolding I don't see a reason to complicate the docs for parity with his docs around scaffolding...

maxffarrell:
@ieedan hmm, Claude did an even worse job splitting these than I thought, I'm going to throw this back in draft and do some cleanup

docs/content/skills.md
maxffarrell: I left it in there, I figured it might be worth explicitly calling out just due to training data inherently having more React & shadcn but isn't really needed for larger models

PR #2680, opened 2026-04-21
fix: slider lose focus after use arrow keys

artur-demidov:
Fix https://github.com/huntabyte/bits-ui/issues/2029


ieedan (approved):
Brilliant thank you!

PR #2681, opened 2026-04-24
Fix : bad render of NativeSelect.Option in dark mode

DrUse1:
Fixing NativeSelect.Option that was rendering badly in dark mode by adding : "bg-[Canvas] text-[CanvasText]" (getting this from shadcn)


ieedan:
Hey @DrUse1 just need to fix the lints on this one and it's good to go!

ieedan (approved):
Thank you!

PR #2691, opened 2026-05-05
fix: demo number input docs homepage wrong height

Lorenx03:
Demo number input in the docs homepage wasn't the same height as the plus and minus buttons.

The fix:
From this:
image

To this:
image


ieedan (approved):
Thank you!

PR #2694, opened 2026-05-09
fix: remove required `page` prop from `Pagination.Previous` and `Pagination.Next`

### LINKED ISSUE #2692: Pagination.Previous and Pagination.Next error
### Describe the bug

This 2 throw this error

```
Type '{}' is not assignable to type 'Omit<{ page: Page; }, "children" | "child"> & { child?: Snippet<[{ props: Record<string, unknown>; }]> | undefined; children?: Snippet<[]> | undefined; style?: any; ref?: HTMLElement | ... 1 more ... | undefined; } & Without<...> & { ...; }'.
  Property 'page' is missing in type '{}' but required in type 'Omit<{ page: Page; }, "children" | "child">'.js(2322)
```

### Reproduction

Use those 2 components.

### Logs

```bash

```

### System Info

```bash
I'm not using npm ;-;
```

### Severity

annoyance

maxffarrell:
Closes #2692

## Problem

`Pagination.Previous` and `Pagination.Next` were typed as `ComponentProps<typeof PaginationLink>`, which inherits `PaginationPrimitive.PageProps` from bits-ui. That type requires a `page` prop — but these components are navigation buttons, not page-number links, so they should never need one.

Users following the documented usage (`<Pagination.Previous />` / `<Pagination.Next />` with no props) got a TypeScript error:

```
Type '{}' is not assignable to type ...
  Property 'page' is missing in type '{}' but required in type 'Omit<{ page: Page; }, ...>'
```

## Fix

Both components now use `PaginationPrimitive.PrevButtonProps` / `PaginationPrimitive.NextButtonProps` and render `PaginationPrimitive.PrevButton` / `PaginationPrimitive.NextButton` directly — the same primitives the existing `PrevButton`/`NextButton` wrappers use. This matches the documented API and eliminates the spurious required prop.

Also updates pagination-demo to use `Pagination.Previous` and `Pagination.Next`


ieedan (approved):
Thank you!

PR #2697, opened 2026-05-12
fix: Sidebar cookie name breaks strict cookie parsers (#2696)

### LINKED ISSUE #2696: Sidebar cookie name breaks strict cookie parsers
### Describe the bug

Currently, the Sidebar component hardcodes the cookie name for storing its state using a colon (`:`):
```typescript
export const SIDEBAR_COOKIE_NAME = "sidebar:state";

```

While this works purely within the browser and `document.cookie`, it violates [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) regarding valid cookie names, which causes issues for strict backend servers. A cookie-name must be a token and a token (defined in [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-2.2) cannot contain separator characters, and the colon (:) is listed as an invalid separator.

If the backend uses a strict HTTP parser (in my case Go's standard [http.ParseCookie](https://pkg.go.dev/net/http#ParseCookie)) it will fail to parse the header upon hitting the colon and will throw an `http: invalid cookie name` error.

### Solution

Maybe change the constant to use a hyphen or underscore:

```typescript
export const SIDEBAR_COOKIE_NAME = "sidebar-state"; // or "sidebar_state"

```

Small note: I realize this is a slight breaking change since existing users will lose their current sidebar state when they update, but it would be great to ensure applications using strict backends don't suffer from random session drops.

### Reproduction

You could verify the issue in a small go (or any other language with a strict cookie parser) example:
```go
package main

import (
	"fmt"
	"net/http"
)

func main() {
	header := "sidebar:state=true; valid_token=12345"
	cookies, err := http.ParseCookie(header)
	
	fmt.Println("Error:", err) 
	fmt.Println("Parsed Cookies:", cookies)
}
```

### Logs

```bash

```

### System Info

```bash
System:
    OS: Linux 7.0 Fedora Linux 44 (Workstation Edition)
    CPU: (20) x64 Intel(R) Core(TM) Ultra 7 265K
    Memory: 19.92 GB / 30.85 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 22.22.2 - /usr/bin/node
    npm: 10.9.7 - /usr/bin/npm
    pnpm: 10.28.2 - ~/.local/bin/pnpm
  Browsers:
    Firefox: 150.0.1
    Firefox Developer Edition: 150.0.1
```

### Severity

annoyance

moaedy:
Closes #2696 ## Problem `SIDEBAR_COOKIE_NAME` was set to `"sidebar:state"`, but a colon is a separator under RFC 6265/2616, so strict server-side cookie parsers (e.g. Go's `http.ParseCookie`) reject the header. The docs site itself was already working around this in `+layout.server.ts` by reading `sidebar_state`. ## Fix Renamed `SIDEBAR_COOKIE_NAME` to `"sidebar_state"` in `docs/src/lib/registry/ui/sidebar/constants.ts`. Running `pnpm build:registry` propagated the change into the generated `docs/static/registry/styles/*/sidebar.json` files.

### DIFF
--- a/docs/src/lib/registry/ui/sidebar/constants.ts
+++ b/docs/src/lib/registry/ui/sidebar/constants.ts
@@ -1,4 +1,4 @@
-export const SIDEBAR_COOKIE_NAME = "sidebar:state";
+export const SIDEBAR_COOKIE_NAME = "sidebar_state";
 export const SIDEBAR_COOKIE_MAX_AGE = 60 * 60 * 24 * 7;
 export const SIDEBAR_WIDTH = "16rem";
 export const SIDEBAR_WIDTH_MOBILE = "18rem";
--- a/docs/static/registry/styles/luma/sidebar.json
+++ b/docs/static/registry/styles/luma/sidebar.json
@@ -18,7 +18,7 @@
 	],
 	"files": [
 		{
-			"content": "export const SIDEBAR_COOKIE_NAME = \"sidebar:state\";\nexport const SIDEBAR_COOKIE_MAX_AGE = 60 * 60 * 24 * 7;\nexport const SIDEBAR_WIDTH = \"16rem\";\nexport const SIDEBAR_WIDTH_MOBILE = \"18rem\";\nexport const SIDEBAR_WIDTH_ICON = \"3rem\";\nexport const SIDEBAR_KEYBOARD_SHORTCUT = \"b\";\n",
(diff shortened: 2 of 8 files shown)


ieedan (approved):
Thank you!

PR #2707, opened 2026-05-17
fix(tabs): use `data-[state=active]` instead of `data-active` in tabs-trigger

### LINKED ISSUE #2706: `tabs-trigger.svelte`: `data-active` should be `data-[state=active]`
### Description

The `tabs-trigger.svelte` component uses Tailwind's `data-active:` shorthand for styling active tab triggers. However, `bits-ui`'s `Tabs.Trigger` sets `data-state="active"` on the active tab — not a boolean `data-active` attribute.

As a result, all active-state styles on `Tabs.Trigger` are silently broken: the active tab gets no background, text color, shadow, or line indicator.

### Affected file

[`docs/src/lib/registry/ui/tabs/tabs-trigger.svelte`](https://github.com/huntabyte/shadcn-svelte/blob/main/docs/src/lib/registry/ui/tabs/tabs-trigger.svelte)

### Current (broken)

```
data-active:bg-background
data-active:text-foreground
group-data-[variant=default]/tabs-list:data-active:shadow-sm
group-data-[variant=line]/tabs-list:data-active:shadow-none
group-data-[variant=line]/tabs-list:data-active:bg-transparent
group-data-[variant=line]/tabs-list:data-active:after:opacity-100
…
```

### Expected (fix)

Replace all `data-active:` with `data-[state=active]:` to match the attribute that `bits-ui` actually sets:

```
data-[state=active]:bg-background
data-[state=active]:text-foreground
group-data-[variant=default]/tabs-list:data-[state=active]:shadow-sm
group-data-[variant=line]/tabs-list:data-[state=active]:shadow-none
group-data-[variant=line]/tabs-list:data-[state=active]:bg-transparent
group-data-[variant=line]/tabs-list:data-[state=active]:after:opacity-100
…
```

### Note

The `doc-tabs-trigger.svelte` [already uses `data-[state=active]` correctly](https://github.com/huntabyte/shadcn-svelte/blob/main/docs/src/lib/components/doc-tabs/doc-tabs-trigger.svelte), so this appears to be an oversight in the registry component only.

Stadly:
bits-ui sets `data-state="active"` on the active tab trigger, not a boolean `data-active` attribute. The Tailwind `data-active:` shorthand therefore never matches, silently breaking all active-state styles (background, text color, shadow, line indicator).

This PR replaces all `data-active:` with `data-[state=active]:` in:

- `docs/src/lib/registry/ui/tabs/tabs-trigger.svelte`
- `docs/src/lib/registry/styles/style-nova.css` (`.cn-tabs-trigger` shadow rules)
- `docs/src/lib/registry/styles/style-vega.css` (`.cn-tabs-trigger` shadow rules)

Note: [`doc-tabs-trigger.svelte`](https://github.com/huntabyte/shadcn-svelte/blob/main/docs/src/lib/components/doc-tabs/doc-tabs-trigger.svelte) already uses `data-[state=active]` correctly, so this appears to be an oversight in the registry component and two style files only.

Closes #2706

### DIFF
--- a/docs/src/lib/registry/styles/style-nova.css
+++ b/docs/src/lib/registry/styles/style-nova.css
@@ -1242,7 +1242,7 @@
 	}
 
 	.cn-tabs-trigger {
-		@apply gap-1.5 rounded-md border border-transparent px-1.5 py-0.5 text-sm font-medium has-data-[icon=inline-end]:pr-1 has-data-[icon=inline-start]:pl-1 group-data-[variant=default]/tabs-list:data-active:shadow-sm group-data-[variant=line]/tabs-list:data-active:shadow-none [&_svg:not([class*='size-'])]:size-4;
+		@apply gap-1.5 rounded-md border border-transparent px-1.5 py-0.5 text-sm font-medium has-data-[icon=inline-end]:pr-1 has-data-[icon=inline-start]:pl-1 group-data-[variant=default]/tabs-list:data-[state=active]:shadow-sm group-data-[variant=line]/tabs-list:data-[state=active]:shadow-none [&_svg:not([class*='size-'])]:size-4;
 	}
 
 	.cn-tabs-content {
--- a/docs/src/lib/registry/styles/style-vega.css
+++ b/docs/src/lib/registry/styles/style-vega.css
@@ -1238,7 +1238,7 @@
 	}
 
 	.cn-tabs-trigger {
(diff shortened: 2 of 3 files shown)


ieedan:
Thanks for the fix @Stadly but if we are going to fix this we will probably go in a different direction since this will make it harder to keep parity with the upstream project

Stadly:
Oh, I had not set `@import "shadcn-svelte/tailwind.css";`. It is not mentioned in the [manual setup guide](https://shadcn-svelte.com/docs/installation/manual).

PR #2709, opened 2026-05-19
docs: drawer missing direction prop

oniSG:
Hey! This is my first contribution, so bear with me ;].
While using the Drawer component I noticed there was no mention of the direction prop in the docs, even though vaul-svelte supports top, right, bottom, and left. The React shadcn/ui docs cover this with a live example, so I figured it'd be useful to have the same here.
                                                                                                                                                                                                                                                                                                                                                                                         What's in this PR:
 - Added a drawer-direction.svelte example component that renders four drawer triggers, one per direction
 - Added a Sides section to the Drawer docs page with a live preview and code block showing how to use the direction prop                                                                                                                                                                                                                                                                                                              

Ill be happy to adjust anything if needed.


ieedan (approved):
Thank you!

PR #2729, opened 2026-06-04
feat(docs): content search

maxffarrell:
Add full-text docs search to the command menu using FlexSearch, providing the content-search behavior of shadcn/ui’s Fumadocs implementation in Svelte. Build the index from Markdown using the docs renderer’s heading IDs and serve it through `/api/search.json`.

Watch content changes during development, escape highlighted snippets, and retain navigation search if the index cannot load.

Validated with `pnpm lint` and `pnpm check`.


.gitattributes
maxffarrell, line 1: @ieedan would this solve the problem of the registry json files distorting diff sizes in PRs?

PR #2753, opened 2026-06-13
fix(docs): missing home in nav & update select registry files from previous PR

maxffarrell:


maxffarrell:
shadcn ui includes ignored sections on mobile

@ieedan super small fix!

ieedan (approved):
Thank you!

PR #2754, opened 2026-06-13
fix(docs): fix create picker submenu contrast

Abdulmumin1:
Fixes unreadable text contrast in the create/customizer picker submenus.

Picker submenu content uses theme-aware popover colors, but picker items forced all descendant text to `text-neutral-100` and used dark hover styles. This made light-mode submenu text and hover states hard to read.

This updates picker item, radio item, and separator styles so submenu content uses theme-aware foreground, accent, and border tokens.

before: 
image

after:
image

it has been really bugging me everytime i visit the docs page.. hope this it not an intended behaviour and this addresses it. ty.

### DIFF
--- a/docs/src/routes/(app)/(layout)/(create)/components/picker/picker-item.svelte
+++ b/docs/src/routes/(app)/(layout)/(create)/components/picker/picker-item.svelte
@@ -18,7 +18,7 @@
 	data-inset={inset}
 	data-variant={variant}
 	class={cn(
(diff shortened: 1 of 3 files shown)


ieedan (approved):
Just need to fix the formatting and this should be good to go! I think these styles were left over from the original implementation shadcn had where this menu never changed colors but ours does so it broke things in light mode. And I never tested it in light mode cause I'm not a physcopath 😉

Abdulmumin1:
> Just need to fix the formatting and this should be good to go! I think these styles were left over from the original implementation shadcn had where this menu never changed colors but ours does so it broke things in light mode. And I never tested it in light mode cause I'm not a physcopath 😉

make sense.. as a physcopath i notice these stuff 🤣

i've amended the commit.

PR #2761, opened 2026-06-17
docs: align docs ui with shadcn

maxffarrell:
Homepage theming, docs layout (visible separator), CLI tab changed to Command, Steps components made visible, etc.


maxffarrell:
@ieedan this one should be quick

PR #2769, opened 2026-06-20
docs: update Astro installation guide

maxffarrell:
## Summary
- add the Astro v6 Vite noExternal setup for Svelte library dependencies
- document the build error this avoids before the global CSS setup step

## Test plan
- pnpm lint


maxffarrell:
need to check vite v8 for changes

PR #2770, opened 2026-06-21
fix(docs): copy menu wrapping

maxffarrell:
- keep docs copy dropdown actions on one line
- give the dropdown enough width for the longest action label


ieedan (approved):
Thank you!

PR #2773, opened 2026-06-21
fix(docs): match css vars

maxffarrell:
## Summary

Align the docs typography/code styling with the current `shadcn/ui` v4 docs implementation.

This ports the relevant global docs styling from `shadcn/ui/apps/v4/app/globals.css` and matches the docs content color scoping used in `shadcn/ui/apps/v4/app/(app)/docs/[[...slug]]/page.tsx`.

## What changed

- Match the upstream docs color tokens in `docs/src/app.css` for both `:root` and `.dark`:
  - light foreground values now use upstream `oklch(0% 0 0)` tokens
  - dark mode keeps upstream `foreground`, `muted-foreground`, `surface-foreground`, and code foreground values
  - related token fixes include `primary`, `destructive`, `destructive-foreground`, `sidebar-foreground`, `sidebar-ring`, and `selection`
- Scope docs page content to `text-foreground dark:text-foreground` instead of `text-neutral-800 dark:text-neutral-300`, matching upstream docs and fixing dimmer dark-mode body/inline-code text.
- Match upstream pretty-code styling:
  - code block radius uses `var(--radius-xl)`
  - code/code span ligatures are disabled like upstream
  - `text` / `plaintext` code blocks reset line padding/display for composition trees, preserving box-drawing connectors
- Match upstream inline-code wrapping by adding `break-words` to the docs inline code rule.
- Preserve Shiki token-based colors for code spans, including light and dark mode.

## Verification

- Compared against current `shadcn/ui` main (`b59f68e`) for the relevant docs globals, MDX code classes, and docs page color scoping.
- Verified `:root` and `.dark` foreground/code color tokens match upstream for the docs surface.
- Ran `pnpm lint`.

### DIFF
--- a/docs/src/app.css
+++ b/docs/src/app.css
@@ -436,7 +436,7 @@
 	}
 
 	.mdsx code:not(pre code) {
-		@apply bg-muted relative rounded-md px-[0.3rem] py-[0.2rem] font-mono text-[0.8rem] outline-none;
+		@apply bg-muted relative rounded-md px-[0.3rem] py-[0.2rem] font-mono text-[0.8rem] break-words outline-none;
 	}
 
 	.super-debug {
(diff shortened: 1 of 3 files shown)


ieedan (approved):
Thank you!

PR #2675, opened 2026-04-21
create docs: chart easing animation and charts page layout updates

maxffarrell:
⚠️ **AI-generated description — will be updated before merge.**

## Summary

- Updates all 70 chart block components: area, bar, line, pie, radar, radial, and tooltip variants
- Adds `charts.ts` config for the charts page and updates `+layout.svelte`
- Adds `defaultBarMotion` and `easing.ts` to chart UI (`src/lib/registry/ui/chart/`)

## Test plan

- [ ] `pnpm --filter docs check` passes with 0 errors
- [ ] `/charts` page renders all chart variants correctly

PR #2774, opened 2026-06-21
docs: charts animations & example chrome

maxffarrell:
## Summary

Reopens the chart docs work from #2675 on the reforked `shadcn-svelte-og` fork, rebased onto `main`.

- Updates the chart block examples across area, bar, line, pie, radar, radial, and tooltip variants.
- Adds chart easing support through `docs/src/lib/registry/ui/chart/easing.ts` and exports it from the chart registry UI entrypoint.
- Refreshes the charts page registry config and removes the extra charts layout wrapper.
- Brings the related `dashboard-01` chart/data-table pieces along with the chart block updates.

## Notes

The original draft targeted `feat/max-doc-updates`; this PR keeps the same two commits of chart work but replays them onto `main` from the reforked `maxffarrell/shadcn-svelte-og` repository.

## Validation

- `pnpm lint`


maxffarrell:
Will cut this down to just docs changes

PR #2777, opened 2026-06-25
feat: add MCP server

### LINKED ISSUE #2333: MCP Docs Server
### Prerequisites

- [x] This feature already exists in [shadcn/ui](https://github.com/shadcn/ui) - if not, it won't be considered here so don't continue with your issue.

### Describe the feature

Hello!
Thanks for this awesome project!
I have a simple but really helpful request...
a MCP server for the documentation.

https://ui.shadcn.com/docs/mcp
This is the real one of shadcn but its not working with shadcn-svelte, so could creating one be an option?
Have a nice day!

maxffarrell:
Add `shadcn-svelte mcp` and `mcp init` for Claude Code, Cursor, VS Code, Codex, and OpenCode, following shadcn/ui’s MCP tools. Supports browsing, searching, viewing examples, and generating add commands for the project’s configured registry.

Includes setup docs and a CLI changeset. Registry-directory and multi-registry features remain outside the current shadcn-svelte registry model.

Closes #2333.

Validated with `pnpm lint` and `pnpm check`.


bnn1:
Has been using this MCP server for a couple of weeks, seems to work fine. Any blockers preventing merge into main?

devharshthakur:
manual code review was the reason. Too many PRs to review, sad to see, but required

AdrianGonz97:
Thanks for all the work so far!! 

I see that we're going to need `@modelcontextprotocol/sdk` for this to work, which is quite a [large package](https://npmgraph.js.org/?q=%40modelcontextprotocol%2Fsdk#select=%40modelcontextprotocol%2Fsdk%401.30.0) (almost 30MB!). 

I wonder if we can separate this out into it's own package and have `shadcn-svelte mcp` effectively be a proxy that simply calls the other package instead (`npx @shadcn-svelte/mcp ...` under the hood), rather than bearing the cost of this dep on all of our users. _Or_ direct people to use `@shadcn-svelte/mcp` directly.

maxffarrell:
> Thanks for all the work so far!! 
> 
> 
> 
> I see that we're going to need `@modelcontextprotocol/sdk` for this to work, which is quite a [large package](https://npmgraph.js.org/?q=%40modelcontextprotocol%2Fsdk#select=%40modelcontextprotocol%2Fsdk%401.30.0) (almost 30MB!). 
> 
> 
> 
> I wonder if we can separate this out into it's own package and have `shadcn-svelte mcp` effectively be a proxy that simply calls the other package instead (`npx @shadcn-svelte/mcp ...` under the hood), rather than bearing the cost of this dep on all of our users. _Or_ direct people to use `@shadcn-svelte/mcp` directly.

I'm not opposed but it seems like this would be a different approach from shadcn

maxffarrell:
@AdrianGonz97 which approach would you prefer? I don't mind making changes

PR #2785, opened 2026-07-02
chart: fix the stale lc-spline-path override

Nic-Polumeyv:
LayerChart dropped the spline-path class in 2.0.0-next.45 when Path was extracted from Spline (techniq/layerchart#659), so the `[&_.lc-spline-path]:opacity-100` utility in chart-container no longer matches anything. That also broke the intent documented above the line. Hovering a stacked area chart dims the other series' boundary lines while their fills stay at full opacity, and multi series line charts dim on hover too. Thanks @MacroYau for catching that removing the utility alone would keep that behavior broken.

Splines render as bare `lc-path` since the extraction, and bare `[&_.lc-path]:opacity-100` would overreach. Pie slices (`lc-arc-line`) and rounded bars (`lc-bar`) also render through Path, so it would kill their hover dimming and leave stacked bars half dimmed, since only the rect-rendered segments would fade. The utility is now `[&_.lc-path:not(.lc-bar):not(.lc-arc-line)]:opacity-100`, which covers exactly what `lc-spline-path` used to.

The style chart.json files embed the same string and were updated to match. The legacy static/registry/chart.json is not touched by the build so I left it as is.


MacroYau:
Hi, I came across this PR when I was diagnosing a bug about unexpected hover dimming of stacked series.

It is expected (as documented in the in-line comment right above your change) that hovering a point or series in a stacked chart should keep all series at full opacity. But the current state of the `ChartContainer` breaks this intent (see the "Area Chart - Interactive" [example](https://shadcn-svelte.com/charts/area#charts) in the official website).

Maybe instead of removing `[&_.lc-spline-path]:opacity-100`, we should change it into `[&_.lc-path]:opacity-100` to restore the intended UX?

Nic-Polumeyv:
You're right, the intent is broken on main and removal alone keeps it that way. Bare `.lc-path` overshoots though. Pie slices (`lc-arc-line`) and rounded bars (`lc-bar`) render through Path too, so it would kill their hover dimming and leave stacked bars half dimmed. Updated the PR to use `[&_.lc-path:not(.lc-bar):not(.lc-arc-line)]:opacity-100` instead, which restores exactly what `lc-spline-path` covered, including line charts which lost the same behavior.

PR #2814, opened 2026-07-20
fix(cli): fall back to running package manager when detected one isn't installed

Saad5400:
### Problem
Running `add` with **npm** and installing dependencies can crash with a "pnpm is not installed" error in a project that never used pnpm.

`detect()` from `package-manager-detector` prioritizes the `packageManager` field in `package.json` (and parent-dir lockfiles) over the project's own `package-lock.json`. Templates commonly ship `"packageManager": "pnpm@x"`, so the CLI resolves pnpm and runs `pnpm add …` — which fails when pnpm isn't installed. Selecting all components makes the dependency list non-empty, so the install command actually runs and the crash surfaces.

### Fix
Probe whether the detected package manager is actually installed; if not, fall back to the package manager running the CLI (`getUserAgent()`, i.e. npm under `npx`), then npm. Projects with a correctly-installed pnpm/yarn/bun are unaffected — the fallback only fires when the detected PM is genuinely absent. The same guard is applied to `syncSvelteKit`.

### Tests
- Unit tests for the new `isPackageManagerInstalled` helper.
- `installDependencies` test asserting the install command falls back to npm when the detected pnpm isn't installed.

### DIFF
--- a/.changeset/pm-fallback-when-not-installed.md
+++ b/.changeset/pm-fallback-when-not-installed.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix(cli): fall back to the running package manager when the detected one (e.g. from a stale `packageManager` field) isn't installed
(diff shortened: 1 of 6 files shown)


Saad5400:
This issue occurred with me when using svelte templates, and Laravel svelte starter template.

ieedan:
IMO I don't think we need this. If your parent repo disagrees with your child repos config then that's just a bug in the repo. Failing is probably the behavior I would expect. I certainly wouldn't want to use the wrong package manager by accident just because I didn't have the correct version of a package manager installed.

PR #2817, opened 2026-07-27
docs: fix broken embla-carousel links in carousel.md

caujasutom:
This fixes 5 broken links to embla-carousel.com in carousel.md. They were all pointing to old URL paths that now return a 404.

Updated the doc and api links at the top of the file, along with the ones in the About, Options, and Plugins sections. They now point to the v8 docs, which matches the version shadcn-svelte actually depends on (embla-carousel-svelte ^8.6.0).


AdrianGonz97 (approved):
thank you!

PR #2822, opened 2026-07-28
fix(cli): resolve wildcard path alias roots

### LINKED ISSUE #2821: CLI rejects root of wildcard path aliases
In a non-SvelteKit Vite project with a `"$lib/*"` tsconfig path, `add` rejects the default `"lib": "$lib"` components.json alias:

```bash
$ pnpm dlx shadcn-svelte@latest add alert
┌   shadcn-svelte  v1.4.2
│
└  Config Error: Invalid import alias found: ("lib": "$lib") in components.json.
   - Import aliases must use existing path aliases defined in your tsconfig.json.
   - See: https://shadcn-svelte.com/docs/installation/manual#configure-path-aliases.
    at ... (file:///<redacted>/node_modules/shadcn-svelte/dist/index.mjs:...)
```

The wildcard is useful here because the app imports files under `$lib`, like `$lib/utils`, but `$lib` itself is not a module. Adding a bare alias would make tsconfig claim that an import works when it does not.

Since `lib` is only used as a destination directory, the CLI should be able to resolve it from the existing wildcard alias.

mustafa0x:
Fixes #2821.

Resolve directory aliases such as `$lib` from an existing `$lib/*` tsconfig path. Includes a focused regression test and patch changeset.

### DIFF
--- a/.changeset/calm-tools-resolve.md
+++ b/.changeset/calm-tools-resolve.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": patch
+---
+
+fix: resolve wildcard path alias roots for components.json directory aliases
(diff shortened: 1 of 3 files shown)


AdrianGonz97 (approved):
thank you!

PR #2832, opened 2026-08-02
fix(docs): honor ?preset= over the stored preset

### LINKED ISSUE #2831: create: shared ?preset= links are ignored, localStorage wins
### Describe the bug

Sent someone a `/create?preset=…` link and they got their own last theme instead of mine. The code in the URL is only ever used as the *initial* value of a `PersistedState`, so once anything is stored under `design-system-preset` the param is dead:

```ts
// docs/src/lib/features/design-system/components/design-system-provider-state.svelte.ts:67-70
this.#preset = new PersistedState<string>(
	"design-system-preset",
	this.#getSearchParam("preset") ?? encodePreset(DEFAULT_PRESET_CONFIG)
);
```

runed's `PersistedState` prefers an existing storage entry over the initial value it was handed (`persisted-state.svelte.js:59-80` in 0.37.1), so the param only ever gets a chance on an empty profile. And it doesn't land there either: the state is constructed during prerender, where `page.url.searchParams` throws and `#getSearchParam` swallows it (`:156-163`, there's already a TODO on that catch). So `b0` is written to storage and the URL is discarded.

That makes the Share dialog and `shareUrl` (`:345-347`) hand out links that don't work for the recipient. `ui.shadcn.com` doesn't have this, it keeps the preset purely in the URL and only stores `shadcn-create-welcome-dialog`, so `?preset=` applies on a cold load.

Workaround is the Open Preset dialog, which goes through the setter and does apply.

### Reproduction

This is the deployed docs app, so no StackBlitz. On shadcn-svelte.com:

1. fresh browser profile, open `https://www.shadcn-svelte.com/create`, hit Shuffle
2. in the same profile open `https://www.shadcn-svelte.com/create?preset=b1FktqnKU`
3. the address bar says `b1FktqnKU`, the sidebar still shows the shuffled theme and `--preset <old code>`

Also on a completely fresh profile going straight to step 2: `localStorage["design-system-preset"]` ends up `"b0"` and the default neutral theme renders.

Same steps on `https://ui.shadcn.com/create?preset=b1x9M8ZeJW` apply the theme.

### System Info

Bug is in the deployed docs site, not in an installed package. Reproduced on Chromium 151.0.7922.34 and Helium, macOS 26.6.

### Severity

annoyance

stickerdaniel:
Fixes #2831

`/create` hands out `?preset=<code>` links through the Share dialog, but the code was only used as the initial value of the `PersistedState` holding the preset. runed prefers an existing localStorage entry over that initial value, so for anyone who had opened the editor before, a shared link did nothing and they kept their own theme. On a cold profile it was a race with the embedded preview iframe, which loads without a query and writes the default under the same key.

This captures the param once and, when it decodes, assigns it after construction so the URL wins. Without a param nothing changes: a returning visitor keeps their stored preset, a first-time visitor gets the default, and an invalid code is ignored rather than persisted.

Verified in a browser against `pnpm dev` and a production build: a shared link overrides a stored preset, applies on a cold profile across repeated runs, no query keeps the stored preset, no query on a cold profile falls back to `b0`, `?preset=not-a-preset` is ignored without page errors, and editing after opening a shared link still round-trips to the URL.

#2755 touches the same constructor. It gives the preview iframe its own storage key but still passes the URL as the initial value only, so the two changes are orthogonal.

### DIFF
--- a/docs/src/lib/features/design-system/components/design-system-provider-state.svelte.ts
+++ b/docs/src/lib/features/design-system/components/design-system-provider-state.svelte.ts
@@ -4,6 +4,7 @@ import { StateHistory, Context, PersistedState } from "runed";
 import {
 	decodePreset,
 	encodePreset,
+	isValidPreset,
 	DEFAULT_PRESET_CONFIG,
 	type PresetConfig,
 	PRESET_BASE_COLOR_KEYS,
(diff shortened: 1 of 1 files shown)


ieedan (approved):
Thank you!

PR #2848, opened 2026-08-16
fix(sheet): use data-[state] variants for open/close animations

### LINKED ISSUE #2815: Latest sheet classes are wrong
### Describe the bug

The `sheet-content.svelte` component generated by the CLI uses Tailwind data-attribute variants `data-open:` and `data-closed:` (e.g. `data-open:animate-in`, `data-closed:animate-out`, `data-[side=right]:data-open:slide-in-from-right-10`) to drive the open/close animations. However, bits-ui's `Dialog.Content` (which `Sheet.Content` wraps) exposes state via a `data-state="open" | "closed"` attribute, not a boolean `data-open` / `data-closed` attribute. Since no element ever carries a literal `data-open` or `data-closed` attribute, these Tailwind rules never match, so `animate-in`/`animate-out`, `fade-in-0`/`fade-out-0`, and the slide-in/out utilities never apply. The sheet now opens and closes instantly with no transition.

The fix is to use `data-[state=open]:` / `data-[state=closed]:` instead of `data-open:` / `data-closed:` throughout the class string (this is how the previous version of the component was written, and it worked correctly).

### Reproduction

1. Install the latest `sheet` component into a new repo via the shadcn-svelte CLI.
2. Open the sheet.

Expected: sheet fades/slides in from the configured side.
Actual: sheet appears instantly with no animation, and closes instantly as well.

### Logs

```bash

```

### System Info

```bash
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 3.07 GB / 15.71 GB
  Binaries:
    Node: 24.17.0 - C:\nvm4w\nodejs\node.EXE
    npm: 11.13.0 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 11.15.0 - C:\Users\Lukeh\AppData\Local\pnpm\pnpm.CMD
    bun: 1.3.14 - C:\Users\Lukeh\.bun\bin\bun.EXE
  Browsers:
    Edge: Chromium (123.0.2420.65)
    Firefox: 150.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
```

### Severity

annoyance

reysilvaa:
Fixes #2815

## Problem

`Sheet.Content` wraps bits-ui's `Dialog.Content`, which exposes its state via `data-state="open" | "closed"`. The class string used `data-open:` / `data-closed:` Tailwind variants, which never match — so the sheet had no enter/exit animation.

## Change

Replace the `data-open:` / `data-closed:` variants with `data-[state=open]:` / `data-[state=closed]:` in `sheet-content.svelte`.

## Verification

- `svelte-check` (docs): 0 errors, 0 warnings
- `oxfmt --check` on the file: pass
- Confirmed against `bits-ui/dist/bits/dialog` that content sets `data-state` via `getDataOpenClosed(...)`

Scope is the single component from the issue; other components with `data-closed:` usages (dropdown-menu, menubar, sidebar) are left untouched.

### DIFF
--- a/docs/src/lib/registry/ui/sheet/sheet-content.svelte
+++ b/docs/src/lib/registry/ui/sheet/sheet-content.svelte
@@ -35,7 +35,7 @@
 		data-slot="sheet-content"
 		data-side={side}
 		class={cn(
-			"cn-sheet-content data-open:animate-in data-open:fade-in-0 data-[side=bottom]:data-open:slide-in-from-bottom-10 data-[side=left]:data-open:slide-in-from-left-10 data-[side=right]:data-open:slide-in-from-right-10 data-[side=top]:data-open:slide-in-from-top-10 data-closed:animate-out data-closed:fade-out-0 data-[side=bottom]:data-closed:slide-out-to-bottom-10 data-[side=left]:data-closed:slide-out-to-left-10 data-[side=right]:data-closed:slide-out-to-right-10 data-[side=top]:data-closed:slide-out-to-top-10",
+			"cn-sheet-content data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[side=bottom]:data-[state=closed]:slide-out-to-bottom-10 data-[side=left]:data-[state=closed]:slide-out-to-left-10 data-[side=right]:data-[state=closed]:slide-out-to-right-10 data-[side=top]:data-[state=closed]:slide-out-to-top-10 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[side=bottom]:data-[state=open]:slide-in-from-bottom-10 data-[side=left]:data-[state=open]:slide-in-from-left-10 data-[side=right]:data-[state=open]:slide-in-from-right-10 data-[side=top]:data-[state=open]:slide-in-from-top-10",
 			className
 		)}
 		{...restProps}


ieedan:
The reason for this is parity with shadcn. These classes are resolved by the shadcn-svelte/tailwind.css import in your global css file. 

I don't agree with it but seeing that they haven't responded to my issue upstream it doesn't look like it's going to change.

reysilvaa:
Thanks for the explanation @ieedan — that makes sense. I see now that `data-open:`/`data-closed:` are intentional for parity with shadcn and are resolved via the `shadcn-svelte/tailwind.css` import, so changing the component classes would break that convention. Closing this was the right call.

To help close out #2815 properly, I posted a triage comment there with the full picture for affected users: the CLI has auto-added `@import "shadcn-svelte/tailwind.css"` on both `add` and `update` since April 2026 (commit `9ff93f8`, #2614), so running `npx shadcn-svelte@latest update` (or `add sheet`) is the fix for anyone whose global CSS is missing the variants — only manual copy-paste or older CLI versions hit the silent break.

One thing I think would genuinely help future upgraders: a short note in the docs (e.g. the upgrade guide or the Sheet page) stating that this import is required in global CSS, especially when copying components by hand. Happy to open a small docs PR for that if you'd like.

PR #2856, opened 2026-08-28
fix(sheet): correct default registry RTL motion

### LINKED ISSUE #2855: Default sheet registry slides from the opposite edge in RTL
The default `sheet` registry item applies explicit RTL overrides on top of the direction-aware `slide-in-from-start` / `slide-in-from-end` utilities.

As a result, horizontal sheets are anchored to one logical edge but animate from the opposite edge in RTL. The base logical utilities already mirror themselves, so the explicit RTL overrides should be removed.

mustafa0x:
Fixes #2855

This follows up on #2563. That PR added explicit RTL slide overrides, but `tw-animate-css@1.4.0` already makes the logical `start` and `end` utilities direction-aware. The additional overrides therefore mirror the animation a second time, making horizontal sheets enter from the opposite edge in a true RTL document.

This removes the redundant overrides from the default sheet registry and lets the logical animation utilities handle both directions.

Tested by parsing the registry JSON and verifying the horizontal motion classes.


ieedan (approved):
Thank you!

Issue #2665, opened 2026-04-19
Cannot install `dashboard-01` block due to `error: zod@catalog: failed to resolve` error


0x113:
### Describe the bug

I'm trying to install `dashboard-01`sample, but I'm running into the following error:
```
■  Failed to install dependencies
│  [0.02ms] ".env"
│  bun add v1.3.12 (700fc117)
│  Resolving dependencies
│  Resolved, downloaded and extracted [67]
│  error: zod@catalog: failed to resolve
│
└  Operation failed.
```

### Reproduction

1. Run `bun x shadcn-svelte@latest add dashboard-01` command.
2. Selection option to ovewrite resources.
3. The command will fail with the error mentioned above.

### Logs

```bash
N/A
```

### System Info

```bash
  System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M4 Pro
    Memory: 1.77 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.8.1 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - ~/node_modules/.bin/yarn
    npm: 9.8.1 - ~/node_modules/.bin/npm
  Browsers:
    Brave Browser: 145.1.87.192
    Chrome: 146.0.7680.165
    Firefox: 109.0.1
    Safari: 26.2
  npmPackages:
    @sveltejs/kit: ^2.57.0 => 2.57.1
    bits-ui: ^2.16.3 => 2.17.3
    shadcn-svelte: ^1.2.7 => 1.2.7
    svelte: ^5.55.2 => 5.55.3
    tailwindcss: ^4.2.2 => 4.2.2
```

### Severity

annoyance

PR #2858, opened 2026-09-01
fix(registry): pin zod instead of catalog specifier in dashboard-01

alalloush:
#2840 re-introduced the pnpm `catalog:` specifier into the published dashboard-01 registry item, regressing #2665 (previously fixed by #2666).

**Repro** (shadcn-svelte 1.5.1, Bun, 2026-08-31):

```
bunx --bun shadcn-svelte@latest add dashboard-01 --yes
# all files scaffold, then:
# error: zod@catalog: failed to resolve
# CLI Error: Operation failed.
```

A consumer project has no matching pnpm workspace catalog, so the specifier cannot resolve. The failure output above is from Bun; exact behavior will vary by package manager.

**Evidence**

- Live: `https://shadcn-svelte.com/registry/styles/vega/dashboard-01.json` → `devDependencies` contains `"zod@catalog:"` (checked all 257 vega registry items on 2026-08-31; dashboard-01 is the only affected item).
- Source: `docs/static/registry/styles/*/dashboard-01.json` plus `docs/package.json` (`"zod": "catalog:"`) at tag `shadcn-svelte@1.5.1` and current main.

**Fix** — same approach as #2666: pin `zod` in `docs/package.json` and the eight per-style `dashboard-01.json` items to the workspace catalog's current value (`^4.4.3`). Lockfile regenerated with `pnpm install --lockfile-only` (pnpm 10.32.1).

Longer term, resolving `catalog:` specifiers during registry generation would prevent the next regression; happy to look into that if you'd prefer it over the pin.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -134,6 +134,6 @@
 		"vaul-svelte": "1.0.0-next.7",
 		"velite": "^0.3.1",
 		"vite": "^8.1.5",
-		"zod": "catalog:"
+		"zod": "^4.4.3"
 	}
 }
--- a/docs/static/registry/styles/luma/dashboard-01.json
+++ b/docs/static/registry/styles/luma/dashboard-01.json
@@ -14,7 +14,7 @@
 		"@dnd-kit-svelte/svelte@^0.1.5",
 		"@dnd-kit/abstract@^0.1.21",
 		"@dnd-kit/helpers@^0.1.21",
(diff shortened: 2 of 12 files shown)


AdrianGonz97 (approved):
thank you!

AdrianGonz97:
for posterity, this regression was introduced in #2824, not #2840

PR #2862, opened 2026-09-01
feat: auto-detect `#lib` subpath import alias during `init`

### LINKED ISSUE #2849: SvelteKit 3 support
### Describe the bug

Unable to add new components after creating a new SvelteKit 3 project and initializing shadcn-svelte

### Reproduction

Create a new SvelteKit 3 project with `npx sv@next create`, minimal template, add Tailwind CSS, and then run `npx shadcn-svelte@latest init`.
Then, running an add command produces this error: 
```bash
npx shadcn-svelte@latest add button
┌   shadcn-svelte  v1.5.0 
│
└  Error: File '$app/tsconfig' not found.
```
This is because the new tsconfig.json in SvelteKit 3 is
```json
{
	"extends": "$app/tsconfig",
	"compilerOptions": {
		"strict": true
	},
	"include": ["src", "vite.config.ts"]
}
```

### Logs

```bash

```

### System Info

```bash
Latest versions of everything.
```

### Severity

blocking an upgrade

AdrianGonz97:
closes #2849 

As SvelteKit 3 now recommends using [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) and no longer include the `$lib` path alias by default, we'll now auto-detect if the project contains the `#lib/*` subpath import during `init`.

I'll update the documentation in a different PR.

### DIFF
--- a/.changeset/tricky-wings-run.md
+++ b/.changeset/tricky-wings-run.md
@@ -1,0 +1,5 @@
+---
+"shadcn-svelte": minor
+---
+
+feat: auto-detect `#lib` subpath imports from `package.json`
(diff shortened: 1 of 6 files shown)


olegpolin:
Didn't fully close #2849 , try the reproduction

PR #2864, opened 2026-09-01
docs: update docs for SvelteKit 3

AdrianGonz97:
Continuation of #2862. This can be merged once SvelteKit 3 officially releases.

### DIFF
--- a/docs/content/cli.md
+++ b/docs/content/cli.md
@@ -20,11 +20,11 @@ You will be asked a few questions to configure `components.json`:
 ```txt showLineNumbers
 Which base color would you like to use? › Slate
 Where is your global CSS file? (this file will be overwritten) › src/routes/layout.css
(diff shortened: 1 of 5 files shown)


E021ntox:
When using `shadcn-svelte`, a large number of extra packages get installed into `package.json`. This litters the project configuration and creates confusion. Original `shadcn/ui` doesn't have this issue. Is there a way to improve this so that only `shadcn-svelte` is referenced in `package.json`? The ideal practice would be like `daisyui`, where only `daisyui` itself needs to be referenced as a single package in `package.json`.

@AdrianGonz97 Thank you for your work!

E021ntox:
Also, the styles of some components are outdated compared to the original version. Please sync them to the latest styles. You can use AI to help you compare the differences.

@huntabyte

PR #2874, opened 2026-09-09
fix(drawer): restore direction layout lost when styles were synced

### LINKED ISSUE #2875: Drawer renders off-screen: direction layout dropped from .cn-drawer-content in #2870
### Describe the bug

The drawer renders only its overlay. `Drawer.Content` ends up `position: fixed` with `inset: auto`, so it falls back to its static position at the end of `<body>` and sits below the page content, off-screen. This affects all eight styles and all four directions.

It is a regression from #2870. That PR synced the style sheets from shadcn/ui, where the drawer's direction layout had moved out of `.cn-drawer-content` and into the component. `style-maia.css` still carries upstream's note about it:

```css
.cn-drawer-content {
	/* Shared surface only. Direction layout is owned by each primitive TSX. */
```

`drawer-content.svelte` still relies on that class for layout, so `inset-x-0`, `bottom-0`, `inset-y-0`, `left-0`/`right-0`, `w-3/4`, `sm:max-w-sm`, `mt-24`/`mb-24` and `max-h-[80vh]` were dropped from every style.

The same sync also removed the direction-based centering from `.cn-drawer-header`, which `drawer-header.svelte` does not supply either.

I went through the rest of that sync to check whether anything else regressed the same way. `.cn-drawer-handle` and `.cn-drawer-overlay` also lost declarations, but both are covered — the component supplies the handle's visibility, and vaul-svelte ships its own fade keyframes. The two `.cn-button-variant-*` changes are intentional restyles (`hover:bg-secondary/80` became a `color-mix` hover), not losses. The drawer is the only component left broken.

PR: #2874

### Reproduction

https://shadcn-svelte.com/docs/components/drawer — click "Open Drawer". The overlay dims the page and the drawer never appears. The Direction example further down the same page behaves identically for top, right, bottom and left.

In the inspector the `[data-slot="drawer-content"]` element is present with `data-state="open"`, but computes to `inset: auto` and sits at its static offset (~3300px down on that page) rather than at the bottom of the viewport.

Also reproducible from a fresh install: `npx shadcn-svelte@latest add drawer` on any style produces a `drawer-content.svelte` whose class string has no direction utilities.

### Logs

```bash
No console or server errors.
```

### System Info

```bash
  System:
    OS: Linux 6.8 Linux Mint 22.1 (Xia)
    CPU: (16) x64 Intel(R) Core(TM) i5-14400
    Memory: 14.10 GB / 31.12 GB
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.23.2
    npm: 10.9.8
    pnpm: 10.32.1
  Browsers:
    Brave Browser: 152.1.94.117
    Chrome: 151.0.7922.169
    Firefox: 152.0.3
  npmPackages:
    @lucide/svelte: ^1.27.0 => 1.27.0
    @sveltejs/kit: ^2.70.1 => 2.70.1
    bits-ui: ^2.16.3 => 2.16.3
    shadcn-svelte: workspace:* => 1.6.1
    svelte: ^5.54.0 => 5.54.0
    tailwindcss: ^4.3.3 => 4.3.3
    vaul-svelte: 1.0.0-next.7 => 1.0.0-next.7
```

### Severity

blocking an upgrade

nureddinhaji:
Fixes #2875

#2870 synced the style sheets from shadcn/ui. Upstream had moved the drawer's direction layout out of `.cn-drawer-content` and into their own component — their `style-maia.css` still documents it:

```css
.cn-drawer-content {
	/* Shared surface only. Direction layout is owned by each primitive TSX. */
```

`drawer-content.svelte` still relied on that class for layout, so the sync silently removed every direction utility across all eight styles: `inset-x-0`, `bottom-0`, `inset-y-0`, `left-0`/`right-0`, `w-3/4`, `sm:max-w-sm`, `mt-24`/`mb-24` and `max-h-[80vh]`. The content stays `position: fixed` with `inset: auto`, so it renders at its static position below the page and only the overlay is visible.

Reproducible on the docs site: https://shadcn-svelte.com/docs/components/drawer — the overlay dims the page and the drawer never appears, for every direction.

The same sync dropped the direction-based centering from `.cn-drawer-header`, which `drawer-header.svelte` doesn't supply either.

`styles/README.md` says those files aren't to be edited here, so this restores both in the components instead, matching what shadcn/ui does.

The restored utilities are the ones `.cn-drawer-content` and `.cn-drawer-header` provided before #2870, and they were identical across all eight styles — nothing style-specific moves into the components.

Verified for all four directions at 1280×900 and 375×667. Also regenerates the eight `static/registry/styles/*/drawer.json`.


ieedan (approved):
Thank you! 

In the future please write your own PR/Issue descriptions. The ones written by AI are super hard to parse. This one as an example probably could've just been:
```md
shadcn moved the positioning styles for drawer content away from the shared styles in the `style-*.css` files back into the base registry components. When #2870 landed this broke the drawer since it no longer had positioning styles. 
```

A lot easier to read than a big block of text that includes a bunch of specific class names and other things that just aren't necessary :)

nureddinhaji:
Thanks for the note, you're absolutely right. I'll follow it.

PR #2471, opened 2025-12-17
feat: create

ieedan:
This PR (finally) brings the create features to shadcn-svelte. That being said we did make some decisions that make this a bit different.

## Deviations
Just a list of choices I have made so far that differ with the original

### 1. Design system persistence for the entire site

Instead of just showing the preview with the design changes we change the entire site. These updates persist to local storage or the url if you are on the /create page. The advantage of this is that we can show what it looks like in every example and block.

### 2. No shadcn-svelte/tailwind.css

shadcn now requires users to have shadcn as a dependency in users tailwind config. I kinda disagree with this as it goes against what makes shadcn so good in the first place so instead we just add the custom variants and classes to the users tailwind config on init.

### 3. Prefer `init` over project templates

While shadcn is supporting project templates at this time we feel that adding support for `init` is a better path forward. In Svelte we have the Svelte CLI which is a powerful tool for project scaffolding and is well maintained and kept up to date with the most recent best practices in Svelte, having project templates doesn't allow us to leverage it in the same way.

## Notes
- We can no longer run registry build on HMR since we need to build 5 different variants and inject styles
- This doesn't bring the CLI all the way up to date with RTL and other features but I want to get this out before shadcn moves the goalpost again. So I will come back to these in a future PR

Fixes #2463
Fixes #2522
Fixes #2508
Fixes #2504
Fixes #2411
Fixes #2441 - Enables `--fail-on-warnings` so that we catch these in the future

PR #2878, opened 2026-09-10
fix(registry): restore max-height on floating content components

### LINKED ISSUE #2877: Regression(Select): floating content components lost
### Describe the bug

`Select.Content` has `overflow-y-auto` but no `max-height`, so it grows to the full height of its item list instead of being capped to the available viewport space. With a long list the popover runs past the bottom of the window.

Nothing scrolls it back. `scrollHeight === clientHeight` on the content, so `overflow-y-auto` never engages and the viewport inside it is never compressed, which also means `SelectScrollUpButton` and `SelectScrollDownButton` can never appear. Items below the fold are unreachable by wheel and by keyboard, and since the content is `position: fixed` the page cannot scroll to them either.

bits-ui publishes the value needed to bound it, on that same element. It is never read:

```
--bits-select-content-available-height: 764px;
```

`max-width` is `none` for the same reason, so long item labels also push the content wider than the window.

**Bisect.** `select-content.svelte`:

| revision | date | `max-h` present |
| --- | --- | --- |
| `0a17af2b` | 2025-06-08 | yes |
| `7ff65b48` | 2025-12-15 | yes |
| `24b3ae4a` (#2471) | 2026-03-18 | no |
| `main` | 2026-09-09 | no |

`dropdown-menu-content.svelte` lost `max-h-(--bits-dropdown-menu-content-available-height)` in the same commit. As of today no floating content component in the `nova` style declares a `max-h`: select, dropdown-menu, popover, context-menu, menubar.

**Expected.** Content capped to the available viewport height and scrolling internally, as it did before #2471.

I intend to submit a PR: restore `max-h-(--bits-<name>-content-available-height)` on the content class of each of those five components. Happy to scope it to `select` alone if you prefer a narrower change.

Image

### Reproduction

https://github.com/imlargo/bug-repro

Open the select. Measured on `[data-slot="select-content"]` at viewport 1280x800:

| | |
| --- | --- |
| rendered size | 1062 x 5600 px |
| computed `max-height` / `max-width` | `none` / `none` |
| `scrollHeight` / `clientHeight` | 5600 / 5600 |
| wheel scroll 3000px | no movement |
| 40x ArrowDown | highlighted item at y=1520, off screen |

### Logs

```bash
No console or server output. The failure is purely layout.
```

### System Info

```bash
System:
    OS: macOS 26.3.1
    CPU: (8) arm64 Apple M3
    Memory: 164.61 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.6.1 - /opt/homebrew/bin/node
    npm: 11.9.0 - /opt/homebrew/bin/npm
    pnpm: 10.29.2 - /opt/homebrew/bin/pnpm
    bun: 1.3.9 - /Users/imlargo/.bun/bin/bun
  Browsers:
    Chrome: 152.0.7977.83
    Safari: 26.3.1
  npmPackages:
    @lucide/svelte: ^1.43.0 => 1.43.0
    @sveltejs/kit: ^2.63.0 => 2.70.3
    bits-ui: ^2.19.2 => 2.19.2
    shadcn-svelte: ^1.6.1 => 1.6.1
    svelte: ^5.56.1 => 5.57.0
    tailwindcss: ^4.3.0 => 4.3.3
```

### Severity

annoyance

imlargo:
Closes #2877.

`Select.Content`, `DropdownMenu.Content` and `ContextMenu.Content` carry `overflow-y-auto` but no `max-height`, so each grows to the full height of its item list instead of being capped to the available viewport space. With a long list the popover runs past the bottom of the window, and nothing scrolls it back: `scrollHeight === clientHeight` on the content, so `overflow-y-auto` never engages, the viewport inside it is never compressed, and the scroll buttons can never appear. The content is `position: fixed`, so the page cannot scroll to the hidden items either.

The three components had this class until #2471, which dropped it while restyling to `nova`:

| revision | date | `max-h` present |
| --- | --- | --- |
| `0a17af2b` | 2025-06-08 | yes |
| `7ff65b48` | 2025-12-15 | yes |
| `24b3ae4a` (#2471) | 2026-03-18 | no |

This PR puts it back, using the value bits-ui already publishes on that same element:

```diff
- z-50 overflow-x-hidden overflow-y-auto
+ z-50 max-h-(--bits-select-content-available-height) overflow-x-hidden overflow-y-auto
```

The class goes inline next to `z-50` rather than into a `cn-*` utility, matching where the components already keep layout concerns such as `w-(--bits-dropdown-menu-anchor-width)`. No style CSS defines `max-height`.

### Scope

Only the three components that regressed. `popover` and `menubar` never declared a `max-h`, so adding one there would be new behavior rather than a restore, and it seemed better to keep this PR to the regression. Happy to widen it if you would rather fix them together.

`max-width` is `none` for the same reason, and long item labels push the content wider than the window. That one also predates #2471, so it is left out here.

### Files

3 source components under `docs/src/lib/registry/ui/`, plus the 24 registry JSON files regenerated by `pnpm build:registry` (3 components across 8 styles). Each generated file differs by one line, the serialized component. This follows the shape of `bc3c2038` (`fix(drawer)`).

No changeset: registry component fixes have not carried one.

### Verification

Repro: a `Select` with 200 items and long labels, viewport 1280x800.

| | before | after |
| --- | --- | --- |
| content size | 1062 x 5600 px | 1062 x 764 px |
| computed `max-height` | `none` | 764px |
| viewport scrolls | no (5600 / 5600) | yes (5600 / 740) |
| wheel scroll 3000px | no movement | scrolls |
| last of 200 items reachable | no | yes |

Checked against a production build served with `vite preview`, in a fresh browser profile with HTTP caching disabled, using the generated `nova/select.json` from this branch rather than the source file. Tailwind emits `max-height:var(--bits-select-content-available-height)` into the built CSS.

On the bits-ui side, `getFloatingContentCSSVars()` sets the variable and the Floating UI `size()` middleware that computes it is not gated behind `avoidCollisions`, unlike `shift()` and `flip()`. The bits-ui docs use this exact class on `Select.Content`.

One note for the reviewer: bits-ui names the variable `--bits-combobox-content-available-height` when `Select.Root` is used as a combobox (`isCombobox`). That does not affect this registry, which builds combobox from `popover` and `command`.


imlargo:
Closing this in favour of #2881.

#2876 landed after I opened this and already restored `max-h` on `dropdown-menu` and `context-menu`, so two of the three changes here are obsolete and the branch conflicts with main. #2881 is branched off current main and scoped to `select`, which is the only one still missing the class.

